PR #11, opened 2025-10-28
Merge Districts v1

instafluff0:
The big one. I'm sorry it's so much code - I've tried to keep it as tight as possible.

Let me know if you have any questions or suggestions for revision and I'm happy to take a look.

### DIFF
--- a/civ_prog_objects.csv
+++ b/civ_prog_objects.csv
@@ -813,7 +813,7 @@ ignore,	   0x4507B0,	0x452860,	0x0,		"Unit_ai_move_offensive_unit",			"void (__f
 ignore,    0x44C340,	0x0,		0x0,		"Unit_ai_eval_bombard_target",			"int (__fastcall *) (Unit * this, int edx, int tile_x, int tile_y, int param_3)"
 ignore,    0x558F70,	0x0,		0x0,		"Leader_is_enemy_unit",				"char (__fastcall *) (Leader * this, int edx, Unit * unit)"
 ignore,	   0x4BFD40,	0x0,		0x0,		"City_get_turns_to_build",			"int (__fastcall *) (City * this, int edx, int order_type, int order_id, char param_3)"
-inlead,    0x4BFF80,	0x0,		0x0,		"City_can_build_improvement",			"bool (__fastcall *) (City * this, int edx, int i_improv, char param_2)"
+inlead,    0x4BFF80,	0x0,		0x0,		"City_can_build_improvement",			"bool (__fastcall *) (City * this, int edx, int i_improv, bool apply_strict_rules)"
 ignore,	   0x4E3D90,	0x0,		0x0,		"Main_Screen_Form_is_unit_hidden_from_player",	"bool (__fastcall *) (Main_Screen_Form * this, int edx, Unit * unit)"
 ignore,	   0x4E8850,	0x4F14E0,	0x4E8910,	"Main_Screen_Form_open_right_click_unit_menu",	"void (__fastcall *) (Main_Screen_Form * this, int edx, int tile_x, int tile_y, int mouse_x, int mouse_y)"
 ignore,	   0x4DBA70,	0x0,		0x0,		"Main_Screen_Form_set_selected_unit",		"void (__fastcall *) (Main_Screen_Form * this, int edx, Unit * unit, bool param_2)"
--- a/injected_code.c
+++ b/injected_code.c
@@ -14599,10 +14599,10 @@ patch_City_can_build_unit (City * this, int edx, int unit_type_id, bool exclude_
 }
 
 bool __fastcall
-patch_City_can_build_improvement (City * this, int edx, int i_improv, bool param_2)
+patch_City_can_build_improvement (City * this, int edx, int i_improv, bool apply_strict_rules)
 {
 	// First defer to the base game's logic
-	bool base = City_can_build_improvement (this, __, i_improv, param_2);
+	bool base = City_can_build_improvement (this, __, i_improv, apply_strict_rules);
 	if (! base) return false;
 	if (! is->current_config.enable_districts) return base;
 
@@ -14620,9 +14620,9 @@ patch_City_can_build_improvement (City * this, int edx, int i_improv, bool param
 				wonder_requires_district = true;
 
 			// Can only build wonders that need districts if an incomplete wonder district exists
-			if (wonder_requires_district &&
-			    ! city_has_wonder_district_with_no_completed_wonder (this))
-				return false;
+				if (wonder_requires_district &&
+				    ! city_has_wonder_district_with_no_completed_wonder (this))
+					return !apply_strict_rules;
 		}
 	}
 
@@ -14632,9 +14632,9 @@ patch_City_can_build_improvement (City * this, int edx, int i_improv, bool param
 	if (! needs_district)
 		return true;
 
-	// Human doesn't have appropriate district but needs one
+	// Human doesn't have appropriate district but needs one; allow relaxed checks so UI can gray entry out
 	if (needs_district && is_human) {
-		return false;
+		return !apply_strict_rules;
 	}
 
 	// Ensure AI has the prereq tech for the district


instafluff0:
Dang, ok thanks for letting me know. Any chance you can share the save files? I will try to spin up a quick game and reproduce on my end as well - I've been doing so many AI simulations I haven't noticed this kind of thing, but it is very likely occurring in plain sight, I'd guess.

Also, for the 4 culture per turn, was that only after immediately reloading the save? On the subsequent turn was it still incorrect? Not sure what to think at all of the holy site reverting but I'll figure it out.

One other thought: you didn't by chance reload after switching to my latest version, did you? If the save didn't have natural wonders (district_id 4) previously but did upon reload, I'm wondering if that somehow bumped up the `district_id`s in an off-by-one kind of error. But that wouldn't explain only that individual tile instance being the wrong type, so that's almost certainly not the issue, I suppose.

Also thanks re: `patch_Main_Screen_Form_handle_key_down`. I will give that a shot shortly.

instafluff0:
One other thing I noticed. The campus here is correctly providing +1 culture, and is in the city's work radius (I have it set to 3), which is why the total culture per turn is 11. The culture icon is not being drawn on the tile though and I can't quite tell why.

image

Could you take a quick look at [patch_City_Form_draw_yields_on_worked_tiles](https://github.com/instafluff0/C3X_Districts/blob/master/injected_code.c#L22139) and [draw_district_yields](https://github.com/instafluff0/C3X_Districts/blob/master/injected_code.c#L21965C1-L21965C21) and see if you can find anything off?

maxpetul:
Here's the save: [Hannibal of the Carthaginians, 110 BC.zip](https://github.com/user-attachments/files/23403836/Hannibal.of.the.Carthaginians.110.BC.zip)
It's not just that one tile that reverted from a campus district back to the holy site it was built over, there are several more tiles like that. You can see that I have a few libraries partially built that can't be continued. I'm pretty sure I didn't fetch changes between saving and reloading that game, but I'm not sure. At least the other two issues are fixed now.

I had a look at `patch_City_Form_draw_yields_on_worked_tiles` and `draw_district_yields` but didn't notice anything off. It's odd because that's the last tile visited in a loop over the work area so it feels like an off-by-one bug, but on closer inspection I don't see any.

instafluff0:
Yeah please see my post update just a sec ago. I think it was b/c Natural Wonders were added in the middle of your game, which changed the district_ids. I've played myself through 2 games lately up to the industrial age and have not seen this issue.

maxpetul:
That makes sense, that would explain too how the worker's ongoing job got changed.

maxpetul:
Some observations from my second test game:
- It's annoying how the game centers on the selected unit when control is pressed to activate the district highlights. That's cost me several misclicks because I didn't expect the camera to move at that time.
- The district highlights sometimes stick around too long. There were many times I had to press control again just to make them go away. I don't know what the logic is for turning the highlights off but I feel like it would be best if it were when the control key is raised. That way the highlights would be something you could check quickly by holding control instead of something you activate by pressing control that then doesn't go away at a predictable or convenient time.
- You've probably noticed this in your own tests, but city growth doesn't always get updated when it needs to be. My cities often show that they are pop capped (size in yellow) or starving (red), or show an incorrect number of turns to grow. This started after I set up a distribution hub providing food.
- One minor bug I encountered is that I went to end a turn, got a disorder warning on my capital, zoomed to it and saw it wasn't actually unhappy, then went on to end the turn without problem. That's never happened to me since I made it so that disorder warning recomputes city happiness before its check, but now there must be something else it needs to update as well related to districts. Do you know what that might be?
- Another minor bug: I had a stack of three workers on a tile, one building a district and the other two idle. When I clicked an idle worker it instead touched the one building the district, asking me whether I wanted to cancel that job. I went back and made sure I clicked an idle worker, and it happened again. Not sure what was going on there, and it might be hard to reproduce.

instafluff0:
Re: automanaging citizens, I just reloaded my saved game with the unmodded vanilla game and it still rioted with the governor on, so I think that particular issue is unrelated to districts and perhaps cities can still riot and I've just been not paying attention.

I'm also experimenting right now with an alternative approach to adding in district & distribution hub yields. I thought about this a bit yesterday and my current approach has been to patch `City_recompute_yields_and_happiness` and [add the district & dist. hub shield, food, and gold bonuses directly](https://github.com/instafluff0/C3X_Districts/blob/master/injected_code.c#L9210-L9217) to `Tiles_Food`, `Tiles_Production`, and `Tiles_Commerce`, then recompute the happiness and so on. My reasoning was that this ensured shields would be subject to corruption (in my testing, dist. hubs seemed too powerful so I thought it better they be subject to this penalty like regular shields).

It seems that monkey'ing with those in the aggregate is causing other issues relating to city happiness and so on that are tricky to track down.

The alternative approach I'm trying is to simply patch `patch_City_calc_tile_yield_at` and, if the city center, add ALL district and dist. hub bonuses through that (while preserving vanilla yields in the city view UI). In my brief testing this seems to work but I need to test quite a bit more.

Do you think this is a more reasonable approach? My thought is that having all additional bonuses go through the city center tile in `patch_City_calc_tile_yield_at` should centralize and simplify district bonuses logic upstream. District tiles themselves remain unworkable directly.

maxpetul:
Alright, I'm finally finished reading through this PR. It's overall excellent work. I'm impressed how quickly you were able to get up and running working with C3X and especially how quickly you were able to implement such a large feature.

I can't add notes to specific lines in injected_code.c because GitHub won't even load it, so I'll just list my observations here. First, the problems:
- `set_up_district_buttons` searches for the automate button so it can copy a couple of fields off it, however the automate button may be disabled by C3X through the disable_worker_automation option. That method should instead search for another button like one of the standard actions. I don't think there's a button that all units always have, but surely no modder would disable all the standard actions on workers.
- It's inefficient for `tile_is_adjacent_to_any_city` to loop over cities instead of neighboring tiles, but that function isn't called anywhere so can be deleted.
- `city_radius_contains_tile` won't give correct results for some tiles in expanded work areas because of the issue of cultural/work area neighbor indices versus regular ones. `Map_compute_neighbor_index` gives you a number that does not line up with how tiles are iterated over an expanded work area. You should use the `is->ni_to_work_radius` array like in `patch_Map_compute_ni_for_work_area`.
- In `load_dynamic_district_configs`, the custom config should be loaded after the scenario config for consistency with the base mod. Same goes for `load_dynamic_wonder_configs`. However, in C3X the custom config is intended to be used for user preferences, which doesn't apply exactly to district configs. Maybe the custom districts config should be renamed instead?
- `find_tile_for_neighborhood_district` should not be and-ing tile coords with 0xFF.
- It's a very minor thing, but in `recompute_distribution_hub_yields`, the problem with using pointer values for breaking ties is that they will change when the game is saved and reloaded. Is it even necessary to have a tie breaker there? How could you ever have two distro hubs on the same tile?
- In `patch_Map_check_city_location`, turning on natural wonders blocks founding over all district instances not just natural wonders.
- `patch_City_add_population` is mostly stubbed out by an early return. The extra code looks like scraps that never got deleted.
- `aerodrome_airlift_usage` is not kept in the save file.

Next, some questions:
- Why the `(long)` casts everywhere?
- When adding distro hub yields to the city's tile, why did you put that in a new patch of `City_calc_tile_yield_at` instead of putting it in `patch_City_calc_tile_yield_at_while_gathering` alongside the logic for yields from generated resources?
- Why are hub bonuses tracked separately for cities in addition to civs?
- Should district_worker_records and related things be cleared when the worker is despawned? That's how I would have done it but maybe you've thought of that and handled it another way.
- Is there some reason you had to build the trade network and recompute city connections in `patch_do_load_game`? The game of course already builds the network after loading the game, does it not do that at a convenient time for you?

Lastly, some remarks. None of these are things I necessarily expect to be changed before we post R26 Preview 1.
- I don't mind the defensive programming. It makes sense because we don't know what kind of guarantees the original code makes, and besides the original code is itself defensive in a lot of ways. However, I think you're being too defensive at times. For example, `is` will never be NULL unless the mod is completely broken so it's no use to check for that in the mod code itself. `tile_at` can never return NULL, only the special null tile. The `FOR_TILES_AROUND` macro will never give an invalid tile and similarly `FOR_CITIES_OF` will never give an invalid city. That's part of the point of using those macros, actually, is that they already include the necessary checks.
- Many of the small strings like "%s requires a Neighborhood to grow", "Volcanic eruption destroys %s!", etc. should be entries in c3x-labels.txt for ease of translation. There is in fact a German translation of the mod though I haven't checked whether it's been updated in several years.
- It's odd that districts have a multiplier on defense bonus instead of adding a bonus of their own. Also, you have the multiplier applied to the tile building defense bonus, but that comes from cities, fortresses, and some other thing related to Tile field D4 that I don't understand. Can fortresses be built on the same tile as districts? Is that all their multiplier is intended to be good for?
- If an AI builds a wonder in a district one tile over its border with you, can you steal the wonder by planting a city next to it, like with resources? I'm guessing not because the wonder improvement is only granted when the wonder is completed.
- A macro like `FOR_WORK_AREA_AROUND` for a given center tile could replace many repetitive loops in the districts logic. I considered adding that back when I implemented the expanded work areas but decided back then that there weren't enough loops of that type to bother.
- For district highlights (`compute_highlighted_worker_tiles_for_districts`), it would make more sense to me if highlights showed only how many cities were in range, not how many are in range and also in range of the selected worker. I.e., the highlight shouldn't depend on the location of the worker. I also think the hash table there is some premature optimization, but that doesn't matter.
- Wouldn't it be simpler to initialize `building_name_to_id` all at once in `load_scenario` instead of in `load_districts_config`?
- District yield icons on the map in the city screen don't appear when zoomed out.

Again, great work. I'll try to fill in the Steam and PCG addresses soon so we can post R26P1 by the end of this week.

maxpetul:
>Ah, good point. I've switched this to use the "Explore" button instead.

Okay, but... explore is like the one standard button I would not have chosen because if you're going to take any of the standard actions away from the worker, explore would be it. The idea there being that you never use workers to explore but accidentally press 'e' when reaching for 'r', so it's convenient to drop the explore ability.

>Huh. Thank you for pointing this out and I may need to rethink this, as after implementing the change and loading a save, my distro hub I was using became unused, I think because (correctly) the distro hub was not in my anchor city's work radius: ... This makes me wonder whether I should rework the anchor city assumption logic, and simply find an anchor city based on shortest proximity, rather than work area.

Yes, that disto hub is in the fourth ring. Maybe I'm missing something but I don't see why anchor cities are necessary. Why not just use the trade network on the hub's tile? The game tracks network connections on each tile, it has to otherwise you wouldn't be able to make tile improvements require resources. Every contiguous net of roads is associated with one city that's accessible through it. That's stored in the Tile_Body.connected_city_ids field, one for each player.

>First though, would this be reasonable logic for city_radius_contains_tile? I'd rather just leverage patch_Map_compute_ni_for_work_area than repeat its logic:

That works.

>Oh interesting, this was a fundamental misunderstanding of mine. However I had also envisioned in C3X Districts to have the custom.* files basically hold user preferences, which could be overridden however if the user happened to load a specific scenario. (e.g., in a typical game the user doesn't like Encampments, so removes those, but if they happen to load a scenario that uses Encampments, they can do so without having to change their custom config files).

The awkward thing here is that the C3X configs contain both special game rules and purely cosmetic settings like the grouping of units on the right click menu. The custom config was intended for those cosmetic things, although of course you can set game rules there too. So loading the custom config after the scenario one makes sense because cosmetic preferences ought to apply to all games regardless of scenario. If players want to edit the C3X game rules, my intention is they'd create a scenario for that.

Districts are different in that the district configs contain only game rules, so there it does make more sense to load the custom config after the default one and before the scenario. However, calling it "custom" is potentially confusing.

>Given my inexperience in writing C, when I was reading up on how to do pointer-to-int casts, I believe I read at some point that it could be an issue if pointers are 64 bits but ints are 32, so to avoid an issue with the compiler it's not a bad idea to convert to a long as an intermediate step. I started doing that so somewhat mindlessly continued. If it's unnecessary or I misunderstood, I can certainly remove these.

In our case there is no point to the intermediate casts to `long`. In general, there's no point in worrying about other platforms, the Civ 3 EXE is what it is. You can count on the platform being 32-bit, x86, Windows.

>The usage of patch_City_calc_tile_yield_at_while_gathering was just unclear to me, as I saw it was called 6 times, I think, and I didn't understand why or the implications. I can certainly move the logic over if that'd be better.

The idea there was to add in the additional yields only when the game was actually gathering tile yields for a city, like to update its economy, and not at other times like in the AI. I don't remember if there was a specific reason for doing it that way, or if I was just being defensive.

>Only cities connected by trade net to the capital should get distro hub bonuses. In war, for example, if I isolated a city and destroyed its roads to the capital, I'd reasonably expect bonuses to be cut off in that case - was my reasoning.

I was wondering why go through the trouble, if that's the rule. Capital-connected cities all recieve one set of benefits and all others receive zero, and the game already stores which cities are connected to the capital, so you only need to track that one set of benefits. Do you have any plans to allow for distro hubs to function on trade networks not including the capital? It could be useful for oversea colonies before Navigation, so they could have their own little distro networks. The game already tracks what's necessary for that to work, after all it works for resources, so I don't expect it would require much effort to add.

>What I was going for was basically being able to replicate the defensive bonuses of fortresses (if defense_bonus_multiplier_pct is set to 150%, for example) and so on, as well as alternatively to penalize defenders if configured that way, etc. I'm not quite sure what you mean by adding a bonus of their own. Can you elaborate?

`get_building_defense_bonus_at` returns a defense bonus, not a total defense multiplier. So on a fortress, it returns 50, same for a city size 6-12 or smaller with walls, and on a metro it returns 100. On empty tiles it returns zero. All you want to do is add the district bonus on top of that, not multiply it in.

>Sure, good idea. Let me know if this is what you had in mind.

I honestly don't know what I had in mind. It doesn't look any simpler that way but I guess it doesn't matter either way.

>Hmm, they are hard to read but I think working for me. Let me know if I'm misunderstanding.

The issue depends on screen resolution. Here's what it looks like for me in fullscreen (3440x1440):
icons_not_appearing
If I run the game in a 1024x768 window then the icons appear when zoomed out too, like they should.

maxpetul:
>Totally makes sense, assuming I captured that correctly. I'll make the change.

Yes, that's right.

>Yeah I did that in the 2nd commit [here](https://github.com/maxpetul/C3X/pull/11/commits/db93dd2fc730fcbae7fee1932ed6a47893f5b55c), though set Neighborhoods at 25. I suppose I could just zero that out and keep it in the config file

Ah, I didn't notice that. You can set neighborhoods to whatever you want but it should be consistent between the config file and C3X.h, so no change needed.

instafluff0:
Alright, I made a separate branch. Here's a somewhat disappointing update, as I thought we were nearly there. [Here's the diff](https://github.com/instafluff0/C3X_Districts/commit/ea2e0623faa2ec588fd951f7698ab946984b708f).

First, in `patch_City_Form_draw_yields_on_worked_tiles` I added in 

```c
if (center_screen_x < 0)
    center_screen_x += p_bic_data->Map.Width;
```

When I zoomed out in full resolution, the district yields still weren't drawn for me. Is this not what you were thinking? Not sure why, but I was more disappointed by the next item...

---

Pre-commit
image

Post-commit
Screenshot 2025-11-13 at 2 54 59 PM

The non-distro hub yields are entirely off, for reasons I can't figure out. The code changes are not dramatic but there is something I've obviously screwed up.

Moreover, if I destroy the distro hub, the adjacent tiles mysteriously all have their yields shown, despite the fact that they are not worked. Clicking one of the shown-but-not-worked tiles done nothing, presumably because no citizen is actually working them.
image

So I'm re-re-reviewing the code to figure out what subtle difference is causing the bugs. The good news, however, is that it's a substantial reduction of code, and connecting the cities to the hub upon completing roads, etc. works like a charm.

Anyway if anything obviously buggy jumps out to you in the diff please let me know, otherwise I'll keep working on it.

instafluff0:
Got it, thanks. And sorry you probably didn't need the play-by-play. It turned out the issues were (1) needing to zero out the district and distro hub yields in `patch_City_calc_tile_yield_at` (thank you!) and (2) legacy anchor city code that was setting `Body.CityAreaID` for distro hub-covered tiles - which honestly I can't recall why I did it that way in the first place. Fixing both of those solved all issues I could see across multiple games (within which I built, destroyed, rebuilt, and cut off the hubs and effects were as expected, finally).

Here are the diffs - [first](https://github.com/maxpetul/C3X/pull/11/commits/522bcd7fac116214e4baa2ed421eb8cd3ff01910) and [second](https://github.com/maxpetul/C3X/pull/11/commits/587b409b0a1ead0f5cc158c871fcb5d5067128e7)

The only downside is removing the anchor city from the distro hub records necessitated updating serialization code, so previous saves won't work (er, they will, but distro hubs won't work as expected).

Last, as you can see I set `center_screen_x += p_bic_data->Map.Width * tile_half_width;` but for some reason I'm not seeing it work on my end. I haven't dug into it though. If I'm being dense or still not getting that right let me know, else it might be easier for you to update after the merge if you have a better sense of why my change wouldn't work.

maxpetul:
Glad you got it figured out now. Though it occurred to me that one disadvantage to having distro hubs not be associated with a nearby city is that it's less natural and predictable how much food they'll get from water tiles because of harbors. I see that you're passing the connected city from the hub's tile to `calc_tile_yield_at`, which means whether or not that city has a harbor determines whether the hub collects an extra point of food from water tiles.

>Last, as you can see I set center_screen_x += p_bic_data->Map.Width * tile_half_width; but for some reason I'm not seeing it work on my end.

That's very odd because that fixed it for me. You could try adding the same thing for Y just in case, and if that doesn't do it, maybe there's another bug causing the same problem.

instafluff0:
> That's very odd because that fixed it for me. You could try adding the same thing for Y just in case, and if that doesn't do it, maybe there's another bug causing the same problem.

Brilliant. That did it! ([diff](https://github.com/maxpetul/C3X/pull/11/commits/2cf351c49bb4c412040ed8888a69ef33cd29a493))

instafluff0:
[More minor bug fixes](https://github.com/maxpetul/C3X/pull/11/commits/e7c5bc5c13b23e02475e766973f105f0d8b1513e). I've been kicking the tires and trying to break things. I think it we are solid now - though please confirm on your end as well, if you would.

instafluff0:
Oh wow, nice find. I just removed the patch. It makes sense and I was aware it was a redundancy to a certain extent, as making the tiles themselves unworkable was doing the actually enforcement. I've just removed the patch and reverted to 'define'. I tested and see no difference, as we'd expect.

And yes, `Art/Districts/1200` and `Art/Districts/*.pcx` should cover it. (the art looked like it worked fine on your end, right?)

I've made a new updated zip download for day/night cycle art that I'll post after we roll out R26P1.

instafluff0:
Just to amend my update earlier: I think we are solid for R26P1. I'll stop making changes unless I see something glaring. The last thing I've added after some playtesting (and realizing it was kind of annoying not knowing what improvements I *could* build if I had the right districts built) was graying out of improvements in the city view if that's the case:

Example: Can build 2 Wonders that don't need wonder district, but can't build Sistine Chapel for that reason.
image

Example: Can't build barracks, temple, library, or Sistine Chapel due to missing districts.
image

I dug into `City::can_build_improvement` and it turned out this was very simple. I hadn't known what the last param was for, but it's basically a strictness check for this purpose. Changing 2 lines in `patch_City_can_build_improvement` with that was all that was needed. ([diff](https://github.com/maxpetul/C3X/pull/11/commits/43757d44a909f788e35c3937a2984d3cca58430e))

PR #13, opened 2025-11-15
Add work area limits by city improvements

Philiquaz:
Been a while since working straight in C, so I definitely did a bodge job on the config parsing part and I'd definitely read through that carefully.
Unsure about any particular perf concerns of checking the work area like this, but I thought it was fun.
Edits and fixes welcome, but seems to work fine on my machine (tm)

Effects of the addition are documented in the config file.
In summary, this lets improvements extend the work area similarly to the work area by culture option. The resultant area is the intersection (ie minimum radius) of the two mechanics. In addition to allowing buildings to set a maximum radius, they may also (or instead, depending on your mod/scenario) contribute a bonus to the radius on top of the base radius.

One thing I considered was adding other pseudo-improvement checks (beyond just "default") and slapping this into an enum (instead of using SHRT_MAX - no idea how safe that is!)
The first thought that came to mind was to use the special class "freshwater" but then I realised I can't just yoink that from the City struct, as I realised in my testing that if you were to set a bonus radius from using an aqueduct... well freshwater cities would never get it. So feel free to extend.

### DIFF
--- a/injected_code.c
+++ b/injected_code.c
@@ -16338,10 +16338,8 @@ grant_existing_district_buildings_to_city (City * city)
 }
 
 //We need to forwards-declare this
-//edx: return address, doesn't appear to always be used (or at least, running it with 0 doesn't seem to crash the game)
-//param_1: whether to also automatically manage citizen professions
 void __fastcall
-patch_City_manage_by_governor (City * this, int edx, bool param_1);
+patch_City_manage_by_governor (City * this, int edx, bool manage_professions);
 
 void __fastcall
 patch_City_add_or_remove_improvement (City * this, int edx, int improv_id, int add, bool param_3)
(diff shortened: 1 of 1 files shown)


Philiquaz:
I probably misremembered whether default short is signed or not in C (or rather whichever compiler settings we happen to be using because this is C after all), so went with max which would have been the same sort of -1.
As for it being short at all... I think I read around and saw other improv_ids being listed as shorts, so I aimed not to introduce any hassle for myself and did the ol copy paste.
This was all done rather spuriously and without much sleep, so I can definitely see some things being a little silly haha

maxpetul:
To remove improvements, the game calls City::add_or_remove_improvement with add=0. There's already a patched version of that method in the mod code. To redistribute workers, calling patch_City_manage_by_governor should work, I don't actually know what you should pass for the parameter there.

I remember when I was testing the expanded city radius I broke cities somehow so badly that redistributing the workers didn't work. There were some workers that weren't working any tiles and couldn't be clicked or reassigned. I don't remember what caused that, hopefully not contracting the work radius.

Philiquaz:
Ah, typical. Wasn't outside my expectations, good to know.
Then I'll leave this cleaned up as is (2nd try being the charm) and wait for any other thoughts or considerations to manifest
Cheers

PR #16, opened 2025-11-23
[WIP] Introducing city placement conditions and more (the overkill solution)

Philiquaz:
WIP because I haven't written all the rule-parsing or tested it yet.

Greatly expands on previous PR #15 by introducing a rule system to define more categories of minimum separation rules other than "any city" and "any foreign city"

Rule system intended to include support for identifying cities by varying criteria, terrain by criterias, resources and units, and so forth. My hope is the "target" types can be reused down the road when defining rules for other gameplay modifications.

Rules can define minimum and maximum counts (with existing min-separation type rules having a maximum of 0, for example), have 3 metrics to mix+match (including euclidean now [uses a 100x multiplier to give configs some precision]) and support for enabling or disabling by researched techs.

Hopefully this creates enough rule conditions to create all sorts of weird and wonderful scenarios. Zones of settlement denial, only settle with grain nearby, and so forth.

Assistance with writing the parsers would be appreciated, since it's rather dull the motivation to finish might be a bit lacking.

### DIFF
--- a/C3X.h
+++ b/C3X.h
@@ -137,6 +137,66 @@ enum perfume_kind {
 	COUNT_PERFUME_KINDS
 };
 
+enum map_target_type {
+	CITY,//flag foreign, flag population, flag culture
+	TERRAIN,//flag food_yield, flag shield_yield, flag commerce_yield
+	RESOURCE,//flag
+	UNIT,//flag foreign, flag hostile
+}
+
+struct map_target_city {
+	int flags;
+	bool foreign;//flag 0
+	bool hostile;//flag 1
+	int population_min;//flag 2
+	int population_max;//flag 3
+	int culture_min;//flag 4
+	int culture_max;//flag 5
+}
+
+struct map_target_terrain {
+	int flags;
+	int type;//flag 0
+	int food_yield_min;//flag 1
+	int food_yield_max;//flag 2
+	int shield_yield_min;//flag 3
+	int shield_yield_max;//flag 4
+	int commerce_yield_min;//flag 5
+	int commerce_yield_max;//flag 6
+}
+
+struct map_target_resource {
+	int flags;
+	int type;//flag 0
+	bool is_visible;//flag 1
+}
+
+struct map_target_unit {
+	int flags;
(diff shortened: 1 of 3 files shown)


maxpetul:
My ultimate plan for this sort of thing, allowing the game rules to be modified in weird or complex ways, is to integrate Lua with the program and then allow rules to be scripted in Lua. I hope to start on that in earnest before too long, after R26 is done if nothing else comes up. So I wouldn't recommend spending effort on complex pseudo-scripting languages for custom rules when that sort of thing will become obsolete soon (I hope).

maxpetul:
I intend to use LuaJIT so the performance of the Lua code itself should be pretty good. There will still be friction at the interface between Lua and C, and to minimize that I intend to call into Lua only when necessary. In other words, it will be possible to patch functions like it is right now, just with a Lua instead of C replacement, and the mod will only redirect calls into Lua if a replacement is defined. There could still be performance issues, though, for example I'm counting on LuaJIT to be able to access the C structs and functions from Lua without much overhead.

I'd like to convert Civ3Conquests.h into something usable from Lua but I'm not sure the best way to go about that. I tried a few years ago to write a Python script to convert it automatically but decided ultimately that it wasn't a good approach. A straightforward translation of the header file wouldn't give a good interface and trying to make it more abstract while translating wasn't really feasible with a script. I'd like to continue using AI this time. I think converting the interface is a good job for an AI since it's a job that's too mindless to interest me but not so mindless that it could be done by a Python script.

My Lua work so far is on the integrate_lua branch. I've built out just enough to port over the disorder warning feature, the Lua code for it is here: https://github.com/maxpetul/C3X/blob/integrate_lua/script.lua. I'm quite happy with it, you can do things like `for citizen in city:Citizens()` to loop over citizens more easily than in C, and `civ3.ShowPopup` simplifies what would be several function calls in C.

There's no shared Ghidra server exactly, the way I've been doing things so far is I send my Ghidra project to anyone who asks. If you want it, email me. There are a lot of methods I've found in the GOG EXE that aren't listed in civ_prog_objects.csv.

PR #17, opened 2025-11-23
Districts R26P2

instafluff0:
Hi, as discussed I've fixed various (mostly minor) bugs, and also added better transparency for the user as to which config files are loaded and being used, as well as a config check on game load that tells the user of possible mismatches.

1. [Ensure user & scenario override preceding district settings](https://github.com/instafluff0/C3X_Districts/commit/1fe1591bffe2801db2018887b9512263edb5a005)
2. [In patch_City_can_build_improvement, move tech check earlier so buildings whose tech prereq is met but district prereq is not are not shown grayed out in city screen](https://github.com/instafluff0/C3X_Districts/commit/a7e74a7daefb6ba712eb7d7ab218f3aba13a338f)
3. [Ensure is_worker checks are consistent regardless of tile unit is on](https://github.com/instafluff0/C3X_Districts/commit/fcf16c023b3b899c52a83bb2c3743de7a81fc937)
4. [Ensure Automate button for worker doesn't crash game](https://github.com/instafluff0/C3X_Districts/commit/459b2e18a65a704bfa4470c048c04bc6d86f53a7)
5. [Ensure AI can't build on natural wonders](https://github.com/instafluff0/C3X_Districts/commit/5144abd7c720153d35b6a58d8835245301e309dc)
6. [Ensure natural wonders can't be placed on starting positions](https://github.com/instafluff0/C3X_Districts/commit/f851e45a390e357ec8e2853522dc05e277cef388)
7. [Add support for natural wonder placement in swamp, hills](https://github.com/instafluff0/C3X_Districts/commit/68eaf73e5cf919c79d6f3ea52596f372dd7345ca)
8. [Ensure natural wonders can be used even if districts disabled](https://github.com/instafluff0/C3X_Districts/commit/395185673d995a7a4873a05fd4c69556d27287af)
9. [Simplify distro hub logic and ensure surrounding tiles are always unworkable, even after conquest](https://github.com/instafluff0/C3X_Districts/commit/e14668dbb3d0f56617e39847193f1722e72eb2b0)
10. [Prevent AI placing distro hubs next to cities; lengthen right-click menu item string for longer district names](https://github.com/instafluff0/C3X_Districts/commit/b0d6b8b27ce47d388b9be0a0548ed569d5a70885)
11. [Ensure scenario config files must be scenario folder; district coordinates file renamed to consistent `scenario.districts.txt`](https://github.com/instafluff0/C3X_Districts/commit/b5927e6b54ac3ab4c139047105f22036da54d6d7)
12. [Add districts config name check on game load to show warning if IDs or names are different](https://github.com/instafluff0/C3X_Districts/commit/3227113f6f23385755f5008737cb988cf73e629c)

The warning looks like this, and verbiage is defined in `c3x-labels.txt`:

Screenshot 2025-11-23 at 8 48 11 AM

I also updated the C3X Info button to include which districts-related files were loaded (though I can't recall which commit):

image

Thanks and let me know if you think anything should be reworked.

### DIFF
--- a/injected_code.c
+++ b/injected_code.c
@@ -23473,7 +23473,7 @@ bool __fastcall
 patch_Tile_has_city_or_district (Tile * this)
 {
 	bool has_city = Tile_has_city (this);
-	if (is->current_config.enable_districts) {
+	if (is->current_config.enable_districts || is->current_config.enable_natural_wonders) {
 		return has_city || (get_district_instance (this) != NULL);
 	}
 	return has_city;


instafluff0:
Great, thanks. I've just updated now.

> The de/serialize_string_with_len methods aren't called anywhere.

Thanks for catching that, [removed now](https://github.com/maxpetul/C3X/pull/17/commits/8b344c3a9132a86da158838b5763743d459970df).

---

> There's an obsolete comment about reading districts config info from a save, starts with "Only load the districts from the scenario file if..."

Huh, I can't find that for some reason. Perhaps I removed it earlier?

---

> The is_scenario variable around config loading doesn't serve any purpose that I can see. It's set based on whether BIQ_get_asset_path can find a path but all the methods it's passed to check that again anyway.

The `is_scenario` variable was getting used in the 3 districts file check functions (the scenario-flavor file would only be looked for if it was true), but it also occurs to me it's possible the user doesn't have a scenario-specific C3X config but still would want to use districts, in which case it'd be better to check for those files independently anyway. So I've [removed the checks now and each always looks for its appropriate scenario config and loads if found](https://github.com/maxpetul/C3X/pull/17/commits/26b5cc8ad68218e596b87542730d7e7e5744e7b2).

---

> I'm not sure about your approach to checking "bic__in_.tmp" to determine whether the game is...

This is a great point, and I couldn't think of a better way to do it. `patch_Map_place_scenario_things` is so obvious and more natural, thanks. [I've moved all the logic to there](https://github.com/maxpetul/C3X/pull/17/commits/8b344c3a9132a86da158838b5763743d459970df).

--- 

I also found that the latest guy in the forum was probably right and it looks like it was indeed possible for a wonder to be rebuilt if a city with that wonder was razed (as opposed to wonder district being destroyed, which was my focus). [I added a check for that as well to prevent that](https://github.com/maxpetul/C3X/pull/17/commits/bf820104eab22333facad7de06b2e497e12fa5fa).

---

Last, on the districts config mismatch error screen I also added the latest districts config file name itself, to simplify debugging for users still further:

image

---

Before release, could you please do me a favor and start from scratch as well as load a scenario or two with my above changes and confirm things look as expected? I'm pretty confident I've gotten everything right and my tests work as expected, but having confirmation would be great. I'm guessing you do this anyway before release, but either way thank you.

instafluff0:
Quick question: I thought I had prevented the AI from building roads through natural wonders with [this commit](https://github.com/instafluff0/C3X_Districts/commit/5144abd7c720153d35b6a58d8835245301e309dc), based on my reading of what it does in the source code. However I'm still seeing the AI occasionally build roads through them (though not other improvements, I think). Can you think of a better way to approach this?

maxpetul:
To block building roads specifically, you could patch `Tile::get_road_bonus`. But to block all worker jobs on natural wonder tiles, it should work to add a condition to `patch_Leader_can_do_worker_job` to that effect.

instafluff0:
Thanks. Let me make a commit for that and follow up. I think I'm done otherwise.

instafluff0:
Perfect, thanks!

PR #21, opened 2026-01-30
Districts R27

instafluff0:
Hey @maxpetul,

I hope things are going well. I've been hard at work on the new additions to districts. As seems to be my MO these days, I set out with a very specific set of goals and a relatively narrow scope. That lasted about ... 2 weeks. As I reworked things and added new features, those in turn inspired me to add more ... and more. I also have gotten a lot of feedback from modders and so on on CivFanatics and did my best to add straightforward features that weren't much of a lift, or seemed worth it.

Anyway, the PR has become quite big (again), for which I can only apologize for asking you to do yet another huge code review. Thank you very much, as always.

After this gets merged and any outstanding bugs get worked out, I plan to go quiet for a while and take a break. At this point I have no intended major additions I want to do to districts - this PR is kind of the kitchen sink. After this I think districts will actually be very mature, in terms of features and robustness. So I can't make any promises, but at this point the only remaining dev I intend to do is adding a seasonal cycle, which will mostly be art anyway.

## Changelog
Here's a reasonably exhaustive list of changelog items:

- 11 new districts: Port, Canal, Bridge, Municipal, Central Rail Hub, Data Center, Energy Grid, Great Wall, Ski Resort, Offshore Extraction Zone, Water Park
- 7 new natural wonders: Wadi Rum, Eyjafjallajokull, Ha Long Bay, Lofoten Skerries, Geirangerfjord, Delicate Arch, Savanna (mostly a desert/tundra theme this time)
- 3 new wonders: Great Lighthouse, Colossus, Hoover Dam
- Add support for maritime districts
- Move Research Lab from Campus to Data Center dependency
- Add Great Wall auto-build around borders event after Great Wall wonder is completed
- Add support for Named Tiles, which shows custom verbiage on a given tile (can be set in-game or via `scenario.districts.txt`)
- Fix bug with districts not rendering on right-click tile detail modal
- Fix bug with workers "building" a natural wonder if user clicks "Move all units of type" button
- Fix bug with custom scenario district art not loading from scenario folder
- Add support for alternative directions of Wonders, based on city vs. wonder tile location
- Add negative district bonuses
- Add custom logic for river- & port-related rendering
- Add granular configuration control over where districts can be built by tile and adjacent tile types
- Add support for districts being dependent on wonders & natural wonders
- Add support for districts enabled by alliances with other civs that can build them
- Add support for restrictions of who can build districts by civ name, trait, culture, government
- Allow multiple advance prerequisites
- Touch up art for Industrial Zones, Newton's University, Holy Site, Bach's Cathedral
- Add support for buildings being buildable if one of multiple possible districts is built
- Add support for districts generating resources
- Add support for districts requiring resources, both in cities or on a tile
- Add optional flag for districts to serve as alternative irrigation source
- Add patches to enable buildings dependent on adjacent water sources to be enabled if water in work area (eg, Ports can be built if city has water in work radius), except for Aqueducts & Coastal Fortresses, which still require adjacency
- Add flag for districts with dependent buildings to be custom tile improvements, which encourages the AI to evaluate whether to build them on each tile, like irrigation and mines
- Add much smarter AI for distribution hubs, with ai_distribution_hub_build_strategy = auto (AI assess civ-wide and city level needs to determine how best to use tiles and distro hubs to build)
- Add more balance distribution hub yields, with amount of shields/food each city receives decreasing as the number of recipient cities grows
- Add fairly sophisticated AI build strategy and analysis of where to place canals and bridges
- Add cap on max distribution hubs relative to number of cities
- Add flag for show_ai_city_location_desirability_if_settler, which automatically shows city desirability over tiles if a settler is selected. This is extremely useful if minimum_city_separation is higher than normal, as it's hard to tell where cities can be settled otherwise
- Add flag for auto_zoom_city_screen_for_large_work_areas, which prevents the city screen from auto-zooming out if the city_work_radius is >= 4 (my personal preference, but off by default)
- Add support for extraterritorial colonies, where colonies can be built in another civ's territory or be built beforehand but not disappear as borders grow, at a relation penalty
- Add happiness_bonus to district yields
- Add display_name, in case name shown to user is different
- Add optional custom height/width & x/y offsets for district art
- Add error messaging to main screen if buildings / techs aren't parsed correctly, like other config options (in R26 this silently fails)
- Distro hub yields on city screen are drawn even if outside city work radius (useful as in R26 they aren't drawn if outside the work radius, even if city is receiving distro hub yields)
- Make tile highlights green, not red - don't kill me, and let me know if you don't like this. I just found green to be more inviting and intuitive in a "deeper green is better" sense
- Replace all redundant, hard to read district tile loops with proper macros
- Add custom logic for air unit position and bombing if using aerodromes. This was necessary as I found that the AI never moved air units otherwise, and turned out the vanilla code explicitly looped over cities & airfields (duh, I suppose), so if cities weren't valid places to move to, the AI never did
- Add support for alternative rendering logic (in districts config, render_strategy = "by-building") for handling districts whose buildings are not dependent on each other. The main use case for this was the Municipal District, but I wanted to generalize it so anyone could do this and we wouldn't need to add custom districts in the future
- Add custom logic for naval units to heal in ports and pillage enemy ports. This was necessary as otherwise, if naval_units_use_port_districts_not_cities = true, the AI never healed ships and never attacked others' ports
- Add custom logic for unit movement on bridges, canals, great wall
- Allow flexible additive bonuses for districts in config based on nearby cities' available buildings and district tile & overlay types
- Allow naval units to pillage (maritime districts)

## Current issues need advice on
I'm making the PR now as (1) I don't anticipate any other major changes to the codebase unless you request it, and (2) I've basically hit a wall on a few items I don't know how to resolve and need your help. Most of these are related to unit movement, which I think you understand much better than I do.

Those are:
1. **Naval units can't actually enter canals**. I've reviewed the vanilla code extensively and tried many different patches, but I can't get it to work. The pathfinder works fine, and correctly calculates the number of turns it would take to move and suggests the unit can pass. But when I actually have a ship try to enter, it just doesn't. Kind of pulling my hair out on this one. Relevant functions are `patch_Unit_can_move_to_adjacent_tile` and `patch_Unit_move_to_adjacent_tile`.
2. **Land units can successfully move on bridges and workers can enter coastal tiles (if workers_can_enter_coast = true), but even if set on a multi-turn path, they wake up every turn**. This is annoying. I suspect this is due to water rules with land units in transports, where they wake up if not sleeping, but again haven't been able to pin down exactly where in the codebase this actually happens.
3. Related to (2), as I said, land units can successfully move on bridges, including handling railroads (yay), but **when they leave a bridge and move to land - even if both tiles have roads/railroads - it still takes a full turn**. I assume this is due to vanilla logic of land units moving from a transport to land, but haven't figured out where this happens or where to patch. 
4. Figuring out how to enable the Pillage command/button for naval units was a huge pain. I initially tried to do so via checks in `patch_Unit_can_perform_command` and `patch_Leader_can_do_worker_job`, but those seemed to have no effect and I think for naval units they perhaps aren't checked at all? Not completely sure. **As a workaround, I patched `patch_Main_GUI_set_up_unit_command_buttons` and put the check there, BUT the code I got working to actually enable the button is ugly** and took a lot of poking around the source to understand the offsets and get the right images. I hate it but couldn't figure out a better way. This works but I would love suggestions for a better way to do it.
5. I was able to figure out how to negatively affect a Leader's attitude toward one another by setting a penalty for each extraterritorial colony, in `patch_Leader_get_attitude_toward`. The relation penalty in the config is called `per_extraterritorial_colony_relation_penalty`. **I couldn't figure out how to actually loop over colonies directly though, so as a workaround loop over every tile on the map**. In the source code I was able to find and add `p_colonies` to civ_prog_objects.csv, but don't quite understand how to loop over it. This is probably really dumb but I could not quite figure out why this works for p_cities and not p_colonies.
6. **The district property `buildable_by_civs` checks civ names by strings**. I *really* wanted to do this by just figuring out the civ_id and indexing that on start/load, but as configs are loaded before civs are decided, I couldn't do my typical approach. If you're ok with this I am too, but just wanted to point that out.

I think the code is relatively clean, and I worked hard to incorporate your feedback from the first districts PR and keep everything tight. Overall I think it is solid. As you review the code though, you will probably come across `align_variant_and_pixel_offsets_with_coastline` and a few other functions which handle rendering maritime district stuff. I'll just say I hated putting specific if/else statements there to try to handle pixel offsets for specific terrain tile sheets and sprites, but it really looked bad without it and I couldn't find any cleaner way. In terms of how far beach/coast/land is on the coastal terrain art, it really is wacky and all over the place. Without handling that, ports and so on often appear way out in the water and look stupid. Anyway long-story-short: I know the code there is ugly and am very open to feedback. But the game looks much better because of it.

## Remaining stuff to do
Like I said the code is basically done, besides refinement based on feedback and thie issues I need help on above. ~~I'm planning to refactor and add a `buildable_on_overlays` property for districts that is distinct from `buildable_on`, and should be done with that in a day or two.~~ (this is done)

There is still some art to finish as well. ZergMazter is doing the art for Central Rail Hub and Municipal District, and probably needs a couple more weeks. I'm working on the light annotations for the new districts as well. So let's wait to actually merge this in until after those are done, but I saw no need to wait for that in order to start code reviews.

Phew! Sorry for the super long explanation. Thank you for your help. I'm very heartened by how many people have reached out and said how happy they are with the new stuff we're adding. A number of people have told me details about custom districts and scenarios they are working on, which makes it all worth it.

### DIFF
--- a/injected_code.c
+++ b/injected_code.c
@@ -24819,17 +24819,20 @@ remove_extra_palaces (City * city, City * excluded_destination)
 	} while (extra_palace_lost >= 0);
 }
 
+// Give a city any completed wonder districts in work radius, if cities_with_mutual_district_receive_wonders is true.
+// This is essentially for cases where the original Wonder-constructing city is lost and a new city is built 
+// that can work the same wonder district tile.
 void
 grant_nearby_wonders_to_city (City * city)
 {
-	// Give a city any completed wonder districts in work radius, if cities_with_mutual_district_receive_wonders is true.
-	// This is essentially for cases where the original Wonder-constructing city is destroyed and a new city is built 
-	// that can work the same wonder district tile.
 	if (! is->current_config.enable_districts ||
 	    ! is->current_config.enable_wonder_districts ||
 	    ! is->current_config.cities_with_mutual_district_receive_wonders ||
 	    (city == NULL))
 		return;
+
+	bool prev_flag = is->sharing_buildings_by_districts_in_progress;
+	is->sharing_buildings_by_districts_in_progress = true;
 
 	FOR_DISTRICTS_AROUND (wai, city->Body.X, city->Body.Y, true) {
 		int x = wai.tile_x, y = wai.tile_y;
@@ -24853,13 +24856,17 @@ grant_nearby_wonders_to_city (City * city)
 		Improvement * improv = &p_bic_data->Improvements[improv_id];
 		if ((improv->Characteristics & ITC_Small_Wonder) != 0) {
 			City * owning_city = get_city_ptr (leaders[city->Body.CivID].Small_Wonders[improv_id]);
-			if (owning_city != NULL)
+			if ((owning_city != NULL) &&
+			    (owning_city->Body.CivID == city->Body.CivID) &&
+			    ! city_radius_contains_tile (owning_city, x, y))
 				continue;
 		}
 
 		// Add the Wonder to the city
 		patch_City_add_or_remove_improvement (city, __, improv_id, 1, false);
 	}
+
+	is->sharing_buildings_by_districts_in_progress = prev_flag;
 }
 
 void
@@ -28756,6 +28763,34 @@ find_city_to_inherit_shared_small_wonder (Leader * leader, int wonder_improv_id)
 }
 
 void
+collect_small_wonders_present_in_city (City * city, int * lost_small_wonders, int * lost_small_wonder_count, int max_lost_small_wonders)
+{
+	if ((city == NULL) ||
+	    (lost_small_wonders == NULL) ||
+	    (lost_small_wonder_count == NULL) ||
+	    (max_lost_small_wonders <= 0))
+		return;
+
+	*lost_small_wonder_count = 0;
+
+	if (! is->current_config.enable_districts ||
+	    ! is->current_config.enable_wonder_districts ||
+	    ! is->current_config.cities_with_mutual_district_receive_wonders)
+		return;
+
+	for (int improv_id = 0; improv_id < p_bic_data->ImprovementsCount; improv_id++) {
+		Improvement * improv = &p_bic_data->Improvements[improv_id];
+		if ((improv->Characteristics & ITC_Small_Wonder) == 0)
+			continue;
+		if (! patch_City_has_improvement (city, __, improv_id, false))
+			continue;
+		if (*lost_small_wonder_count >= max_lost_small_wonders)
+			break;
+		lost_small_wonders[(*lost_small_wonder_count)++] = improv_id;
+	}
+}
+
+void
 reassign_shared_small_wonder_owners_after_city_loss (Leader * leader, int const * lost_small_wonders, int lost_small_wonder_count)
 {
 	if ((! is->current_config.enable_districts) ||
@@ -28789,24 +28824,11 @@ patch_Leader_do_capture_city (Leader * this, int edx, City * city, bool involunt
 {
 	Leader * previous_owner = &leaders[city->Body.CivID];
 	int lost_small_wonders[32];
-	int lost_small_wonder_count = 0;
+	int lost_small_wonder_count;
 	
 	// Record which small wonders were physically present in the city before capture so any
 	// post-capture ownership repair only touches wonders actually affected.
-	if (is->current_config.enable_districts &&
-	    is->current_config.enable_wonder_districts &&
-	    is->current_config.cities_with_mutual_district_receive_wonders) {
-		for (int improv_id = 0; improv_id < p_bic_data->ImprovementsCount; improv_id++) {
-			Improvement * improv = &p_bic_data->Improvements[improv_id];
-			if ((improv->Characteristics & ITC_Small_Wonder) == 0)
-				continue;
-			if (! patch_City_has_improvement (city, __, improv_id, false))
-				continue;
-			if (lost_small_wonder_count >= ARRAY_LEN (lost_small_wonders))
-				break;
-			lost_small_wonders[lost_small_wonder_count++] = improv_id;
-		}
-	}
+	collect_small_wonders_present_in_city (city, lost_small_wonders, &lost_small_wonder_count, ARRAY_LEN (lost_small_wonders));
 
 	is->currently_capturing_city = city;
 	on_lose_city (previous_owner, city, converted ? CLR_CONVERTED : (involuntary ? CLR_CONQUERED : CLR_TRADED));
@@ -28829,8 +28851,16 @@ patch_Leader_do_capture_city (Leader * this, int edx, City * city, bool involunt
 void __fastcall
 patch_City_raze (City * this, int edx, int civ_id_responsible, bool checking_elimination)
 {
-	on_lose_city (&leaders[this->Body.CivID], this, CLR_DESTROYED);
+	Leader * previous_owner = &leaders[this->Body.CivID];
+	int lost_small_wonders[32];
+	int lost_small_wonder_count;
+
+	collect_small_wonders_present_in_city (this, lost_small_wonders, &lost_small_wonder_count, ARRAY_LEN (lost_small_wonders));
+	on_lose_city (previous_owner, this, CLR_DESTROYED);
 	City_raze (this, __, civ_id_responsible, checking_elimination);
+
+	if (lost_small_wonder_count > 0)
+		reassign_shared_small_wonder_owners_after_city_loss (previous_owner, lost_small_wonders, lost_small_wonder_count);
 
 	// Delete the extra improvement bits records for this city
 	City_Improvements * improv_lists[2] = {&this->Body.Improvements_1, &this->Body.Improvements_2};


instafluff0:
Sorry, I know I keep making changes here and there. Mostly these are just renaming districts config fields or adding a few others as I think of use cases the current config surface wouldn't allow. None of them modify core flow, etc.

instafluff0:
> The colonies variable is an object of type Base_List. You had the address set to one of its fields; I switched it to the object itself.

Got it, thank you! I took a longer look at it and understand what you mean now. I updated the loop in `patch_Leader_get_attitude_toward` accordingly (https://github.com/instafluff0/C3X_Districts/commit/c2febaa7aead5f09085e86a183736b85f9ad9bc6)

> Why did you specify temp_ui_strs as an array of 5 elements? I have it set to 10 in Ghidra but I don't know that's necessarily correct.

It was very arbitrary, if I recall correctly. I think I did a search in Ghidra for places it was used, and the highest magic number/literal index I saw was 4. I wasn't sure of a better approach so just used that. I've updated to 10 now as well, at least to align what we have in Ghidra vs. civ_prog_objects (https://github.com/instafluff0/C3X_Districts/commit/a40139df2157e0c54570cc242a1c75cd7d4778b2).

> For Map::is_near_river and similar functions, you can just call them through the vtable. That's more convenient than defining them since that way you (or I) don't have to fill in the various addresses. I made that change.

Ah - that makes perfect sense. I should have thought of it. Thanks for doing that.

> Have you figured out a name for the function at 0x5D3AB0 or any of its callers? Do you know what those functions are about? I'm curious how you decided to replace the call to Tile::has_colony inside it.

It's been a bit since I worked on this, but if I recall correctly I checked in Ghidra for `Tile::has_colony` calls, as I knew that the territory expansion function check(s) almost certainly had to use that. `FUN_005d3ab0` had a number of checks on outposts and good huts as well, so it seemed likely to me it was related to border expansion (though I kind of got lucky, I don't remember checking too many other functions). Based on my read of the code, I had a hunch that if I simply canceled out the `Tile::has_colony` check, they wouldn't get wiped. 

Anyway, I haven't been properly updating Ghidra on my end (given that we're always going to be out of sync, it just didn't really seem worth it, but maybe that's me being lazy). Taking another look now, maybe reasonable name for it might be `Map::handle_tile_border_flip_effects (Map *map, int tile_x, int tile_y)`.

instafluff0:
By the way, I had resolved merge conflicts earlier but they are out of sync again. Git seems quite unhappy with me and I can't figure out a clean way to resolve issues with my master branch and yours. It's a pain but I may need to close this PR, make a branch, merge the branch and your master, then open a PR from that - unless you can think of a better way to do it.

maxpetul:
>this is my fault for making such a huge PR and I feel terrible

It's not a big deal, don't feel bad about it. But, yes, more manageable PRs would be preferred.

>It was very arbitrary, if I recall correctly. I think I did a search in Ghidra for places it was used, and the highest magic number/literal index I saw was 4.

Alright. I'm not surprised that most of those strings aren't used. Let's hope I had a good reason to set the length to 10 in the first place.

>Anyway, I haven't been properly updating Ghidra on my end (given that we're always going to be out of sync, it just didn't really seem worth it, but maybe that's me being lazy). Taking another look now, maybe a reasonable name for it might be `Map::handle_tile_border_expansion_effects (Map *map, int tile_x, int tile_y)`.

I could send you updated versions of the Ghidra project if you'd like. Though I haven't done all that much reverse engineering work since the version I sent you about six months ago, I've just identified some functions here and there, so a current version wouldn't necessarily be useful. It may even be detrimental depending on how many functions you've filled in for yourself.

I agree `handle_tile_border_expansion_effects` looks like the right name for that method. By the way, that method actually has a third `civ_id` stack parameter. You can tell because of the `in_stack_...` variable, that generally means the signature is wrong, there's a large stack allocation too large for the decompiler, or the decompiler is confused by some vtable calls, either those are defined wrong or there are just too many. It's easy to confirm how many stack parameters a `thiscall` method has by checking one of its return instructions. In this case, it's `ret 0xC`, meaning it pops 12 bytes off the stack, corresponding to 12/4 = 3 stack parameters.

>I have a new branch at https://github.com/instafluff0/C3X_Districts/tree/r27-merge that is directly merge-able. I can switch to that if you want. Somedays git just kills me.

Whatever's easiest for you is fine. Unfortunately I can't offer any advice on resolving merge conflicts, all I can offer is my sincere sympathy. Programmer hell is fixing merge conflicts for all eternity.

instafluff0:
> I could send you updated versions of the Ghidra project if you'd like. 

It's ok, I think if both you and I have made mostly modest changes to it, let's hold off for now. Most of the reverse-engineering I do ends up in civ_prog_objects anyway.

> Whatever's easiest for you is fine. Unfortunately I can't offer any advice on resolving merge conflicts, all I can offer is my sincere sympathy. Programmer hell is fixing merge conflicts for all eternity.

Ok, thanks, and completely agree. Let's do this: for now, while you're reviewing the current PR, I'll hold off on closing it and opening the new one to avoid confusion. After we get through everything and issue with my `master` is resolved, I'll make a proper branch from it anew, then close this PR and open one for that. 

> By the way, that method actually has a third civ_id stack parameter. You can tell because of the in_stack_... variable, that generally means the signature is wrong, there's a large stack allocation too large for the decompiler, or the decompiler is confused by some vtable calls...

Got it, thanks! Also - and sorry to bring up a separate item, but on the topic of Ghidra and stack param issues... I've made a bit of a breakthrough for animated resources and terrain in R28. It (+ Seasonal Cycle) are mostly done, and will be far smaller in terms of changes, and each feature a separate PR. 

I'm running into trouble with one function that I'd like your very quick help on, when you have a sec: `0x5DA5E0`, which I'm calling `Tile_spawn_animated_effect` `void (__fastcall *) (Tile * this, int edx, AnimatedEffect effect, int tile_x, int tile_y, bool randomize_start_frame)`. I have no issues for several turns (including running custom effects, below), but it seems that when I get a volcanic eruption / civil disorder, the game crashes with "Access violation executing location 0x00000007".  This only happens when I have an inlead on `0x5DA5E0`. I strongly suspect a stack corruption issue but don't see anything odd in Ghidra from what I can tell.

Any ideas? And again, this is in R28, months away, so backburner right now. However looks pretty cool.

test

maxpetul:
Alright, here's some feedback on PR 21, finally. Sorry this is taking so long, I'm mostly done reading through the code at this point. Here are the big things I've noticed so far:

1. Regarding `patch_Main_Screen_Form_handle_right_click_on_tile`: This method interferes with shift + right click on cities to choose production. When you attempt to do that in game, you get the regular right click menu instead. Also, if you select "Change Production" from that menu, you get the production chooser menu but there's an option at the end to name the tile which does nothing if clicked.

2. Re `tri_next`: The iterator over tile rings does not work properly. For each ring, it iterates tiles up to and including that ring, not just the ones in the ring. Notice how `start_ni` is never used in this method, instead it starts each ring over at neighbor index -1. I wrote a little code to double check this in game too, here, drop this in `patch_Map_Renderer_m19`:
```
	init_tile_highlights ();
	Leader * player = &leaders[p_main_screen_form->Player_CivID];
	City * capital = get_city_ptr (player->CapitalID);
	if (capital != NULL) {
		int ring = 3;
		FOR_TILE_RINGS_AROUND (tri, capital->Body.X, capital->Body.Y, &ring, 1)
			if (tri.tile_x == tile_x && tri.tile_y == tile_y) {
				Sprite_draw_on_map (&is->tile_highlights[COUNT_TILE_HIGHLIGHTS/2], __, this, pixel_x, pixel_y, 1, 1, 1, 0);
				break;
			}
	}
```
It highlights each tile that's included in an iterator over ring 3 around the capital, but notice that all tiles within that ring are highlighted too, including the capital itself.

3. Re `leader_has_pact_ally_district_access`: Shouldn't that check peace treaty _and_ MPP?

4. When you're checking whether or not to show a message to the player about their civ, you should check if the player in question is the UI controller, not just any human player, so the code works in multiplayer. This concerns these functions: `reduce_city_population_due_to_lost_neighborhood`, `remove_dependent_buildings_for_district`, and `patch_Unit_attack_tile`.

5. Re `is_district_command`: `district_id` may be used without having been initialized

6. Re `patch_City_get_building_defense_bonus`: `disable_great_wall_city_defense_bonus` won't work if `optimize_improvement_loops` is turned off.

maxpetul:
I have not looked into those other items yet because I wanted to understand the code first. I'll get to them soon.

maxpetul:
Looks good except one thing. In `patch_City_get_building_defense_bonus`, the logic to cancel the great wall's defense bonus is inside the branch `is->current_config.optimize_improvement_loops` so if that config option is off it will never run and instead base game logic will. You should modify that condition to check for the optimize setting or `cancel_great_wall_boost`, running the custom logic if there's anything that might need it. `combat_defense_improvs` gets initialized in any case so there's no harm in running the custom logic even when the optimize flag is off. Only downside is that turning on the great wall setting will change performance slightly by effectively turning on the improvement loop optimization for that case, but I can't imagine anyone would notice. It's not worth worrying about.

I've finally finished reading through injected_code.c! What an epic it's become, 36000 lines. I have more notes but I've already shared the important things. I'll go over them once more to decide what of the rest is worth bothering you with.

instafluff0:
> Is it really necessary to handle settler highlighting in both patch_Main_Screen_Form_set_selected_unit and patch_Unit_select? I would think the first is sufficient.

https://github.com/maxpetul/C3X/pull/21/changes/7bd81e83877c674814458c53d8081e2ad5971706 - Good call, I think I had originally started with `patch_Unit_select` then added `patch_Main_Screen_Form_set_selected_unit`, without realizing the former was redundant.

---

> In patch_Unit_can_perform_command, in the branch for mining a tile with a district, you shouldn't hardcode which terrain types can be mined since that could change in mods. The proper thing to do there is check if the tile would give any extra shields with a mine using Tile::get_mining_bonus.

https://github.com/maxpetul/C3X/pull/21/commits/a6069872c8b50664623574e3bdb852f62011141f - Great, done.

---

> tile_suitable_for_district redundantly checks district resources; that's already checked by can_build_district_on_tile.

https://github.com/maxpetul/C3X/pull/21/commits/be19d43574d67ee6390cb73f371df5478e19e5e6 - Redundant resource check removed.

---

> district_tech_prereqs maps tech IDs to district IDs, but what if multiple districts have the same tech prereq? It's not a problem right now since this table is not used anywhere.

https://github.com/maxpetul/C3X/pull/21/commits/a5de6fa5f20f0aea10c4efe13359b97615d45729 and https://github.com/maxpetul/C3X/pull/21/changes/4ccd9f8a69cd91110ee413bd7426131186409297 - Riiighttt, gah this was stupid. I think it's a relic either from R26 or my initial R27 thrashing, completely redundant due to `district_infos[].advance_prereq_ids` which does the real work. Removed now.

--- 

> In patch_City_spawn_unit_if_done, what's the point of switching production to a defensive land unit for missing aerodrome/port but then not spawning anything anyway? Seems like it would just annoy the player to have to switch production back or lose shields.

https://github.com/maxpetul/C3X/pull/21/commits/7bbd1e85a9cd046a2935653e39bf0bac25e9d478 - Yes, agreed. I changed this so `patch_City_spawn_unit_if_done` no longer switches human production when an aerodrome/port is missing; it just blocks the spawn and shows the warning, leaving the player’s chosen build in place. The fallback switch is still kept for AI cities, since for them this path is acting as a safety net rather than a UI prompt.

---

> The district_id param is not used in tile_coords_has_city_with_building_in_district_radius.

https://github.com/maxpetul/C3X/pull/21/commits/7cd4f49894e166aa4549ad9b2cd6758b73a53ad5 - Removed.

--- 

> wonder_allows_river is a strange name, what does it mean for a wonder to "allow" a river? Should that be requires_river? Same for district_allows_river but that's not actually called anywhere.

https://github.com/maxpetul/C3X/pull/21/commits/3168b0d74e02f33d8217ee3a39a53c679fae6463 - Yeah that was dumb. I think I was trying to align it with what I had initially been calling the district config-related fields. Renamed.

--- 

> In find_city_by_name, it does not seem helpful to log the name of every city checked. Also this function is not called anywhere.

https://github.com/maxpetul/C3X/pull/21/commits/3138656930cd5202b015ddf902a1827249a7645b - Ugh, thank you, removed.

---

> The comment above load_scenario_districts_from_file, "Attempts the scenario_path first, then scenario_config_path", is obsolete. I don't see any loading from any scenario config path and don't even know what that is.

https://github.com/maxpetul/C3X/pull/21/commits/cbe162c3aa519be1dcdcde09bc0824de6dcc1a2b - That comment has gone the way of the dinosaurs.

--- 

> In patch_City_Form_draw_yields_on_worked_tiles, the condition "Skip distribution hubs if the feature is not enabled" is redundant since that loop skips distro hubs in any case.

https://github.com/maxpetul/C3X/pull/21/commits/fc391161b6fcdbc23ae0f2689cf8ac4f18c1e43b - Nice find, removed now.

---

> There are two junk lines at the start of canal_has_same_sea_isthmus.

> Last and least, the most trifling of complaints: lookup_pending_building_order should be renamed look_up_... because "lookup" is a noun not a verb.

https://github.com/maxpetul/C3X/pull/21/commits/14022b95d9eff86ddc97ff2e4188cd5807cf654c - Removed and renamed.

instafluff0:
Phew! Thank you again. I think that covers nearly everything?

I'll start back on trying to figure out the worker-wake-up-on-coastal-tile-issue*. Also FYI I will be out from the 2nd week of April for 2 weeks. I can communicate a bit but won't be able to code much. I'll try to get everything done I can in the meantime so things don't get held up anymore due my overly huge updates.

*Worst comes to worst I think this could just be in R27 Preview 1 as a known issue, but I'll defer to you.

instafluff0:
> Sure thing, I hope I'm not annoying you with my complaints about the update.

Not in the least. I seriously appreciate working with you and how meticulous and thoughtful you are. I need to up my game and review better. Also makes me think I should experiment with having LLMs do code reviews before I push, though I suspect they'd flag architecture and syntax much more consistently than unintended game implications.

---

> How strange that the base logic in Leader::has_resource_for_job_at loops over multiple cities to check if any connected one has the resource.

Right, you put it this way and I agree. I'm happy to revisit this at a later point too, given how unnecessary it is. But at least it is consistent.

---

> I have a few more concerns related to resources. First, a very minor thing, the comment on resource_prereqs in the default districts config is out of date now.

https://github.com/maxpetul/C3X/pull/21/commits/410ecd2a309ee256aafc9d659bae40ca9c6c01d3 - Good call, updated now.

---

> Second, city_can_build_district is now less reliable. Just because the city itself has access to the resources needed by the district doesn't mean the tiles around it do because there might not be connecting roads. I hope that doesn't cause a problem for the AI. It's probably fine as is, I'm just remarking.

Agreed, though in practice I don't believe this should be a problem. Ultimately for the AI, `tile_suitable_for_district` is going to be what is actually used to place the district, and that'll use the correct resource availability logic. So as far as I can tell it should be ok.

---

> Third, I'd like the logic to be reorganized a bit. I'd like the check for nearby bonus resources moved out of city_has_resource_r. I think district_resource_prereqs_met_r is a better place for it...

https://github.com/maxpetul/C3X/pull/21/commits/410ecd2a309ee256aafc9d659bae40ca9c6c01d3 - Ok, sounds good.

> Fourth, in patch_Trade_Net_recompute_resources, it's a mistake now to check district_generates_resource_for_civ before adding a resource_tile for a district because at that point resource access has not been computed so if the district requires any input resources, those won't be present.

https://github.com/maxpetul/C3X/pull/21/commits/c0d4ce9abf5eea449ca23153956c03b1dfee9406 - Ok, that makes sense and I believe I've got. I changed `patch_Trade_Net_recompute_resources` to stop filtering district resource tiles through `district_generates_resource_for_civ` during the enqueue phase.

The recompute pass now adds a district resource_tile for any completed, owned, non-local district that has a generated resource and passes the tech-side check via `district_can_generate_resource`. Then actual input-resource validation is deferred to `patch_Tile_get_visible_resource_when_recomputing`, which was already doing the right late check with `district_generates_resource_for_civ`.

Let me know if this is what you had in mind and I've captured this correctly. You're right, it's getting pretty complex.

---

> I'm confused about this now. Originally, if a civ lost a city with a small wonder, the logic was to destroy the wonder district containing that small wonder. I said it would make more sense to keep the district as long as there were other cities sharing the small wonder, because otherwise those cities would be left with a wonder that doesn't actually exist on the map. You changed it so that all those cities lose the wonder building, so now if a civ has a small wonder district whose wonder is shared among many cities, if any of those cities gets captured, they all lose the wonder and the district is destroyed. That's a harsh rule in my opinion, but it's fine if you want it to work like that. However that doesn't match what you described to me. I don't see how capturing a city could cause duplication of a small wonder. I'm pretty sure the game won't let you capture small wonders in any case.

Sorry, I was trying to get through the issues and didn't properly explain this one or respond.

So last point first, Virote_Considon was the first to point out that the game - at least via districts - [allowed the capture of a Small Wonder, and afterward had 2 Forbidden Palaces](https://forums.civfanatics.com/threads/c3x-districts.700590/post-16894222). I confirmed this after testing of my own.

_"I captured the Aztec city of Teayo, which had built the Forbidden Palace. The wonder district was intact and I recieved a message telling me I had built the Forbidden Palace. I now have two seperate Forbidden Palaces - one on Teayo, and my pre-existing one in Ta-Tu. Might it be a good idea when capturing Small Wonders with Wonder Districts to check that the wonder hasn't already been built elsewhere by the invading civ?"_

So that happens and may be the source of confusion? In my view, it makes more sense to just destroy the duplicate small wonder. The additional complication of course is that multiple cities may be sharing the newly duplicate small wonder, though it occurs to me now that those would probably each be conquered in turn and thus the code I added is probably unneeded.

Anyway, before I make any further changes here, let me know if this makes sense or I'm misunderstanding.

instafluff0:
> If you're missing anything then I'm missing it too because that looks alright to me. I'll just point out that it would have been cleaner to reassign the owner recorded in Small_Wonders away from the city being lost instead of letting the game clear the owner and filling it back in afterward. You could have put that in on_lose_city so it's done in one place instead of two, and you wouldn't have needed the limit of 32.

Ugh, you're right, that makes sense. Not to kick the can down the road, but if you're alright with this as-is for the time being, I'd like to revisit it later when we have bandwidth instead of refactoring again now. Because...

> How are we looking for a release this weekend? ... Then of course there's the issue of the merge conflicts. I hope that's not a nightmare.

Exactly. I will work on this tonight and tomorrow morning and make a new branch from my master, merge changes with yours, then make another PR (which should have no merge conflicts). When I do so I'll also do as much testing as I can in the time I have, especially with trying out a mod or two again with R27 changes, and also ports/canals/bridges/etc. There's just so much.

On that note, the burden for testing should be mostly on me with districts-related changes of course, but would you mind just doing a real quick set of QC turns with districts after we merge, before release? I'm guessing you do this anyway, and I have no specific concerns, but just to be safe. 

And I haven't figured out the workers-waking-up-each-turn-on-coastal-tiles issue yet. At this point I think it will have to be a known issue with Preview 1 unless you've found anything.

PR #31, opened 2026-04-27
Customizable Unit Counter System & Civ4-Style Best Defender Selection

bingyu-893:
Hi @maxpetul, 

I really love your C3X mod; it’s made playing Civ 3 so much easier and more enjoyable. This has inspired me to develop some new features for C3X, so I’ve had a go at creating a feature: **a configurable unit counter (rock-paper-scissors) system**, alongside an optional **Civilization 4 style "Best Defender" selection logic** for stack combat.

Here is the overall structure of the code:

### 1. Configuration & Toggles (in `default.c3x_config.ini`)
* `enable_unit_counters`: Global toggle. If `false`, the entire system is bypassed.
* `unit_group`: Allows grouping unit types for easier rule matching (syntax matches `building_prereqs_for_units`).
* `counter_rule`: The core ruleset. Supports modifiers for both attacker and defender (e.g., `self_atk_pct`, `enemy_def_pct`). The logic dictates that the first unit listed in the rule is always `self`, and the second is `enemy`, applying bidirectionally regardless of who initiates the attack.
* `use_civ4_style_best_defender`: When enabled (requires `enable_unit_counters`), melee combat will dynamically select the defender that is hardest to defeat based on post-counter combat odds, aligning both the actual combat and the top unit displayed on the map stack.

### 2. Data Structures (`C3X.h`)
* Added `struct unit_counter_group` (Name + array of `UnitTypeID`).
* Added `struct counter_rule` for granular matching:
  * Matches specific unit IDs, `UCM_ANY (*)`, or `UCM_GROUP`.
  * Supports environmental conditions: `terrain_type`, `only_in_city`, `district_id`, and `ignore_terrain` (negates defender's terrain bonuses).

### 3. Core Algorithm
* Implemented `apply_counter_rules(...)` to handle modifier calculation.
* It initializes `aa` (attacker multiplier) and `dd` (defender multiplier) to `100`. It then iterates through all configured `counter_rules` and applies cumulative percentage multipliers for all matching criteria (chained multiplication for multiple valid rules).
* Sets the `ignore_terrain` flag if any matching rule mandates it.

### 4. Combat Context & Engine Hooks
* Introduced `counter_combat_ctx` to store active modifiers (`attacker_atk_pct`, `defender_def_pct`, `ignore_terrain`).
* Hooked `patch_Unit_get_attack_strength` and `patch_Unit_get_defense_strength` to return the modified `base * pct / 100` when the context is active for the current attacker/defender.
* Hooked `patch_Fighter_get_odds_for_main_combat_loop`. Because the original `Fighter_get_combat_odds` reads raw `UnitType.Attack / Defence`, the implementation temporarily alters the `UnitType` table stats based on `apply_counter_rules`, runs the probe, and restores them. 

### 5. Civ4 Best Defender Logic & UI Synchronization
* **Target Selection:** `find_civ4_best_defender_against` enumerates valid defenders in a tile, simulates `apply_counter_rules`, and invokes `Fighter_get_combat_odds`. Selects the candidate with the highest defense odds. In case of a tie, falls back to `Fighter_prefer_first_defender_1` (comparing modified effective defense) to maintain vanilla "cheapest unit defends first" behavior.
* **UI/Combat Sync:** Hooked `patch_Main_Screen_Form_find_visible_unit` to ensure the top unit of the stack matches the predicted best defender. Additionally, hooked `patch_Fighter_fight` to override the engine's pre-selected defender when attacking from an adjacent tile, completely eliminating the "UI shows one unit, combat engages another" mismatch.

~~### 6. Out of Scope
This system currently does not alter target selection for Bombardment, Defensive Bombardment, or Interception (yet).~~

### 6. Bombardment Bonuses (Updated)
* **Update:** Bombardment bonuses are now supported and implemented. The system can now correctly apply modifiers for bombardment actions, excepts Interception.
---

> **Disclaimer:** > As I am still familiarizing myself with this level of C engine hooking, I used AI assistance heavily for the code generation. If you spot any unconventional logic or areas that need optimization, please bear with me—I would really appreciate your feedback and guidance on improving it!


maxpetul:
Hello. Thanks for the contribution. I glanced over it and it looks alright except for a few issues:

- Your commit 466a838 touches every line in injected_code.c and a couple of other files. You must have changed the line endings from DOS-style to Unix-style. That's a problem because it interferes with git blame. Please fix this and make sure your changes match the existing line ending format.
- The district option for counter_rules only works for special districts, that should be mentioned in the config at least. It would be nice if it could work for dynamic districts too, though I'm guessing it doesn't work that way because dynamic districts are loaded after the main mod config so their names couldn't be recognized.
- It should be noted in the config that the valid values for terrain names are the same as for the buildable_on field in the districts config. Those are not the same as what's specified in the BIQ, in particular there are some extra terrains that are recognized like "snow-forest", and only English names are recognized.

maxpetul:
The support for dynamic districts looks good. However, whatever you did to change the line endings to CRLF didn't work; the endings in injected_code.c and other files are still LF. You'll know the line endings are fixed when GitHub no longer shows that this PR touches every line in the files. For example, if you go to the "files changed" tab, for injected_code.c, it shows +37,115 and -36,558, indicating that this PR changers every line in the file. Those numbers should be much smaller showing only your changes. Same applies to civ_prog_objects.csv, default.c3x_config.ini, and C3X.h.

bingyu-893:
Hi @maxpetul,

Just wanted to give you a quick update! I've fixed the line ending issue you pointed out earlier. Additionally, I've implemented a new feature: players can now customize extra combat bonuses based on a unit's experience level directly within the `counter_rule`. 
For my next step, I'm planning to add a feature that displays the combat odds directly in the game. However, I'm not entirely sure how to approach adding or drawing text on the game's UI. Could you give me some pointers or advice on how to hook into the UI to display this kind of information? Any guidance would be greatly appreciated!

maxpetul:
Hello again.
>I've fixed the line ending issue you pointed out earlier.

Thank you, the line ending issue is fixed now. However that has revealed some other issues with this PR, it's undoing some recent commits. For example, looking at the changes to injected_code.c around `wai_init_cities`, it's undoing the `_iter_count` that I inserted to fix a crash. It's also undoing the `max_contiguous_canal_districts` checks that instafluff inserted, the "If the palace has been set as a small wonder..." fix, and many other recent changes.

>For my next step, I'm planning to add a feature that displays the combat odds directly in the game... Could you give me some pointers or advice on how to hook into the UI to display this kind of information?

Are you planning to add that to this PR or start a new one? I'll leave it up to you but I'd prefer if you finished this one off first then started a new one for the combat odds display. As for advice, unfortunately I don't know where you'd hook into the UI, I'd have to look around for that. It also depends how you want to display the odds. I'm imaging a box that appears in one corner of the map display when the player hovers over a target, like in Civ 4. So you'd want to hook into the mouse hover event for Main_Screen_Form, but again, I haven't located that yet. Other than that, you'd probably want to draw a background for the box, that would be `PCX_Image_read_file` to load it and `PCX_Image_draw_onto` to draw it onto the Main_Screen_Form's canvas. Then to add text you'd simply use `PCX_Image_draw_text`, there are a lot of examples of that already in the mod code. I experimented with something like this a few years ago, I'll dig up the branch and see if it's useful.

bingyu-893:
> However that has revealed some other issues with this PR, it's undoing some recent commits.

Thanks for catching that! I will fetch the latest changes from the upstream branch and merge them into mine. Since my new features are mostly isolated to the counter system, there shouldn't be any functional conflicts. I'll get this sorted out soon.

> I'll leave it up to you but I'd prefer if you finished this one off first then started a new one for the combat odds display.

I agree. Splitting this into two PRs is definitely a much better idea for easier management. I'll focus on polishing and finishing up this PR first. 

>  you'd probably want to draw a background for the box, that would be PCX_Image_read_file to load it and PCX_Image_draw_onto to draw it onto the Main_Screen_Form's canvas. Then to add text you'd simply use PCX_Image_draw_text, there are a lot of examples of that already in the mod code.

Thank you for the pointers on the UI side, I will research these methods once I start on the new PR.

maxpetul:
I had a look at that old branch I mentioned. It's probably not going to be useful since back then I was attempting to add a new form to the game with its own widgets such as buttons. That's overkill if all you want to do is display some text.

However, I did some poking around today exploring how the display of combat odds might be implemented. It's actually a feature I've had on my mind for years but have never gotten around to doing myself, so I'm eager to see it done. To capture which tile the mouse is hovering over, you'd want to hook the 28th method (m27) in the Main Screen Form vtable, which I've named `Main_Screen_Form::process_mouse_hover`. That method is called whenever the user moves the mouse on the main map display in-game or over the main menu. Here's a small example of tracking which tile the mouse is hovering over: https://github.com/maxpetul/C3X/commit/70fc4dc697564cb6d0992ffcba260c6f32dfb6b1

I'm still not sure what's the best way to draw the box with some text in it onto the screen. Drawing to the main screen's canvas in `process_mouse_hover` sort of works but it's not consistent with the rest of the drawing logic so whatever is drawn that way is likely to get overwritten. It might work simply to hook `Main_Screen_Form::m22_Draw`, I'll try that next. If that doesn't work, I'll have to look into how the game draws the box to the right that shows passengers in the currently selected unit (if any) and imitate that.

maxpetul:
`Main_Screen_Form::m22_Draw` doesn't actually draw anything, it only sets a dirty flag in the Animator object. That flag is consumed by `Animator::update`, which is where the actual drawing happens. So if you want to add something to the screen, that's the place to do it. Here's another example: https://github.com/maxpetul/C3X/commit/3dea8cad7a584e9cb20b20efa2f3c68967fe5c48. It displays the coords of the tile the mouse is hovering over in the top left of the screen under the menu buttons. Hope this helps.

maxpetul:
I experimented with this PR in game and found a few issues. There's one big issue in that combat animations are messed up and the unit shown on the tile when you attack it is not necessarily the one you end up fighting. My test was that I created a scenario where the player can use cavalry to attack a tile with an enemy rifleman and guerilla. The rifleman was modded to have +7 bonus HP and the guerilla to have -2 bonus HP and +1 defense. The test was to make sure that the game selected the rifleman as the defender since it's more likely to win combat although the guerilla is more likely to win a single round, when use_civ4_style_best_defender is on. What I discovered is that the game will show the guerilla as the top defender but when you attack you end up fighting the rifleman. And another bug, the game will continue showing the guerilla while combat is going on, it will be just standing there while the rifleman fights underneath, not shown.

Then there are the smaller issues:
- In `parse_unit_counter_group`, the call to `skip_punctuation` for a comma at the end of the outer while loop should not be there. The format for the unit_group option should expect commas between groups, not separating units listed within each group. `read_recognizables` already expects commas separating each of the 'recognizables' that it parses, so inner commas conflict with that. Units within each group should be separated by spaces, as is done in the building_prereqs_for_units option. In other words, valid settings for unit_group should look like this: `[Subs: Submarine "Nuclear Submarine", AntiAir: Flak "Mobile SAM"]`.
- Civ 4 style defender selection wouldn't work for army members because there's a separate function that picks which member fights each round of combat for an army. I've called it `Unit::select_army_member_for_combat` and it's already got an inlead on it.
- This is a very minor thing but inside `load_config`, you've placed the branches to handle "unit_group" and "counter_rule" in the wrong place. Notice they're under a comment that says "if key was previously misspelled". They should instead be under the comment that says "if key is for something special". I would place them at the end of that block underneath the branch for "aircraft_victory_animation".

It's not clear to me what the purpose of the `use_civ4_style_best_defender` option is, anyway. If it's intended to make defender selection logic aware of the unit counter rules, as the comment in the config claims, why even have an option for that? Why would anyone turn it off? I think anyone who's using the counter rules would want them to apply to defender selection, so might as well have it active all the time without an option. However, the logic in `find_civ4_best_defender_against` does additional things like skip over invisible defenders to "avoid revealing them". I don't remember if that's how Civ 4 works, but it's a strange rule to have in Civ 3, in my opinion. If you want an option for that, that's fine, but it shouldn't be tangled in with applying counter rules; it should be possible to apply the counter rules to defender selection without changing anything else about it. On a related note, I wonder why you didn't patch `Fighter::prefer_first_defender_1` to modify how defenders are chosen. It seems to me that would be simpler and more reliable than modifying `find_visible_unit` and `fight`, but maybe I'm missing something.

bingyu-893:
Thank you for the detailed testing and feedback! The issues you brought up, especially regarding the UI and combat animation mismatch, are exactly what I've been trying to wrap my head around these past few days. I also completely missed the interaction with Armies (`select_army_member_for_combat`), so I really appreciate you pointing that out.

Unfortunately, my schedule is quite packed at the moment, so realistically it might take me until late May to get all these bugs fixed and the underlying logic refactored as you suggested. 

Regardless, thank you so much for all the guidance. I will get back to working on this PR as soon as I have the time!

maxpetul:
You're welcome, and thanks again for the contribution. Finish off the PR whenever you have the time, there is no rush.

bingyu-893:
Hi @maxpetul,

I managed to find some free time over the last couple of days and actually got these issues fixed. It turned out to be not as difficult as I initially thought once I dug into it. 
While I was at it, I also went ahead and implemented bombardment bonuses for units. I've playtested it across a few games, and everything seems to be running well, no unexpected behaviors or animation glitches this time. 
As for Interception, I haven't implemented it yet since it doesn't seem to cleanly reuse the existing logic. I figure it's best to hold off on that until I have a bit more free time to properly think through how to approach it.

maxpetul:
The line endings got messed up again. I fixed them so they don't interfere with the diff. I had Codex write a Python script to do that, it's called `fix_line_endings.py` and it makes sure that all the line endings in each file follow the original style. That might be useful for you.

This PR still undoes several unrelated changes that are part of the master branch. Merging master in doesn't fix it because those changes got baked in somehow and were covered up by the line endings issue. Just browse through this PR's file changes for injected_code.c and you'll see that it removes, for example:
- the `_iter_count` limit on the `FOR_AERODROME_AROUND` macro
- the `max_contiguous_canal_districts <= 0` checks
- the `!is->current_config.enable_canal_districts` in `can_build_district_on_tile`

maxpetul:
Ok, I had another look at this and found a couple of small issues:
- The entry for `auto_zoom_city_screen_for_large_work_areas` in `boolean_config_options` has been duplicated.
- In `patch_Fighter_damage_by_db_in_main_loop`, you could call `patch_Animator_play_one_shot_unit_animation` instead of setting the display override then calling the base method. That's a bit shorter and it's generally preferred to call the patched methods anywhere outside of the methods themselves.

Have you tested this PR thoroughly, including things like the options related to terrain? I've done a few very basic tests and it seems to work, but I haven't tested anything related to terrain, bombardment, armies, districts, or AI. I'm reluctant to merge because the code still feels overly complicated to me. For example, when I asked you about modifying `prefer_first_defender_1`, I was imagining that you would incorporate the counter rules into the strengths passed to that function then pass them along to the original function. That way the game's original rules about defender selection would apply. I see that instead you reimplemented that logic by calculating the defender's chances to win combat, but your logic is not equivalent to the base game's as it doesn't consider the king flag. In general, I prefer to use the base game logic whenever possible instead of reimplementing it. (By the way, if you want to modify the game's rules for how defenders are chosen, that's fine, just that should be a separate option from the counter rules and in fact could be a different PR.)

I don't have a clear idea of how I'd prefer this code to work instead, so I'm not insisting you change it. If you feel this is the most sensible approach then please go ahead and argue for that. I'm wondering whether it's necessary to modify `prefer_first_defender_1` significantly or even at all, or if you could use the base game's logic. What if instead you simply relied on your existing patches to `get_attack_strength` and `get_defense_strength` and had them return strengths adjusted for a hypothetical attacker or defender unit when those are available? That would be similar to how it already works with the `counter_combat_ctx`. Then you could rely on the original logic for selecting a defender. It would run as normal except use strengths adjusted for a potential attacker. The only problem I can imagine with that is you'd need to incorporate terrain defense bonuses when choosing a defender because those might change depending on the counter rules, but then take care not to double count the terrain effects when computing the actual combat odds. Does that make sense or am I missing something? Sorry if that's how you had it originally before I asked you to change it, I hope we're not going in circles.

bingyu-893:
Thanks, this makes perfect sense. I agree that the current defender-selection code has drifted a bit too far into reimplementing the base-game logic. It started as a workaround while I was chasing the unit animation/display issues, but I agree that relying on the original logic wherever possible is the better direction.

I’ll go ahead and fix the two minor issues you mentioned first. Regarding `prefer_first_defender_1`, your suggestion is definitely the better direction. I’ll simplify that patch so it only folds the counter modifiers into the effective strengths before passing them to the original `Fighter_prefer_first_defender_1`, letting the base game handle the final decision as usual.

I’ve also run tests on terrain-related counter rules, bombardment, armies, and districts, and everything looks solid. I checked the AI side as well: the AI’s attack evaluation and decision-making logic remain unchanged by `enable_unit_counters`. The counters only kick in during the actual combat resolution and defender selection phases once combat is initiated.

On the broader `get_attack_strength` / `get_defense_strength` idea: I think it makes sense conceptually, and it aligns closely with what the current `counter_combat_ctx` does during the actual combat odds calculation. For this PR, however, I’m leaning toward the lighter touch of simply adjusting the strengths at the `prefer_first_defender_1` call site and then delegating to the original function. This achieves the same primary benefit with less global state/context plumbing, and it significantly reduces the risk of accidentally affecting unrelated strength queries. If this still feels too much like a special-case after the simplification, I'd be happy to revisit the broader context-based approach.

bingyu-893:
Hi @maxpetul,

I’ve now reworked the defender selection as you suggested. The counter modifiers are now folded into the effective defense strengths before being passed to the original `Fighter_prefer_first_defender_1`, so the base game still cleanly handles king priority and the remaining tie-breakers.

I’ve also fixed the duplicated config entry, updated the defensive-bombard animation code to call the patched animation method, removed the ineffective "Requires Escort" special case, merged the latest master changes, and corrected the line endings. The compile test passes smoothly, and I’ve successfully playtested terrain rules, bombardment, armies, districts, animations, and AI-controlled combat.

While inspecting the original defender-selection function, I noticed something interesting: when effective defense strengths are equal, the game considers factors like attack and bombard strength, but not shield cost. This leads to situations where a unit with a higher shield cost might defend before a cheaper one if they share the same defense value, simply because the expensive unit happens to have a lower attack strength.

To address this, I added a disabled-by-default option called `prefer_less_expensive_defenders`. It compares shield cost right after effective defense strength and king priority, but before the base game’s remaining tie-breakers. I kept this deliberately independent of `enable_unit_counters`, as it fundamentally tweaks vanilla defender-selection rules rather than implementing counter logic. 

Even though this isn't the original vanilla behavior, sacrificing the cheaper of two equally capable defenders seems to make much more logical sense to me. Would you prefer to keep this as a sperate option, or should I combine it with `unit_counter`?

One last thing: You mentioned in #43 that you have a Ghidra project for Civ 3. If you are comfortable with it, would you be willing to share a copy with me? Having access to your established function names, types, and analysis would be incredibly helpful for my future development and would greatly reduce the risk of me misinterpreting the original game logic.

Thanks again for all the detailed feedback and guidance!

maxpetul:
>One last thing: You mentioned in https://github.com/maxpetul/C3X/pull/43 that you have a Ghidra project for Civ 3. If you are comfortable with it, would you be willing to share a copy with me?

Sure, I'll message you about this on CivFanatics.

maxpetul:
Looking better. Yes, I like having `prefer_less_expensive_defenders` as a separate config option.

Now I'm wondering about your additions to `patch_Unit_select_army_member_for_combat`. For one thing, they don't behave like the base game. In the base game, the strongest unit in an army isn't necessarily the one selected to defend or attack because the army might have run out of HP for that unit. As the army loses HP, the top units become unable to fight and that method selects others further down the list. Your modifications cause the method to return the strongest unit in all cases.

The easiest solution here might be to delete the code you've added to `patch_Unit_select_army_member_for_combat`. The base method relies on `get_attack_strength`, `get_defense_strength`, and `prefer_first_defender`, which have all been modified to be aware of the counter rules, so wouldn't the base method work for the counter rules too? I gave that a quick try and it appeared to work. I commented out the additions to `patch_Unit_select_army_member_for_combat` then attacked an army containing a warrior and spearman with a cavalry unit while warriors were set to have 300% defense against cavalry in the counter rules, and saw that the warrior was chosen to defend above the spearman. So it looks like the counter rules are correctly accounted for there, can you think of any reason they wouldn't be?

Unfortunately there's another issues with armies. Their combat animations are messed up. When combat starts, the HP bar and the little army guy disappear and, when I let the game select different units for combat, the units face the wrong way for a moment when switching.

bingyu-893:
Thanks for your review. I’ve been looking into the issue you mentioned over the past few days, and my conclusions are as follows:

**HP rotation / always returning the strongest:** Fixed — the current version sets a small temporary context (army + defender) and then calls the base selector unchanged, so all of its HP-band allocation and tie-breaking still runs. The context only affects get_attack_strength.

**Defensive selection & Attacker asymmetry:** Fully agree, and that's already what the code does — defensive selection is left entirely to the base method, since prefer_first_defender is counter-aware. That's why your warrior-vs-cavalry test works. I removed my defensive override for this reason.

For the attacker side however, I don't think this works through the base method. The base attacker selection ranks members by get_attack_strength(member) and compares them with the attacker-preference function at 0x4A1280 — which, unlike prefer_first_defender, isn't patched to know about the counter rules.

Additionally, get_attack_strength only applies counter adjustments when a counter context is active: counter_combat_ctx is set only inside patch_Fighter_get_odds_for_main_combat_loop during the actual fight, not during member selection. So with the attack override deleted, the base method picks the strongest base attacker and ignores attack-side counter rules.

The minimal fix I landed on is to make get_attack_strength counter-aware during selection via that temporary context, then call the base selector unchanged. If you'd rather keep it symmetric with the defense path, the alternative is to patch the 0x4A1280 attacker-preference comparator the same way prefer_first_defender is patched and drop the override entirely. Let me know which you prefer.

**Army combat animations:** Fixed. The override now points at the containing army while setting army_top_defender_id to the actual combatant so the right sprite renders, and I orient the shown member toward its opponent to remove the facing flicker on switch.

maxpetul:
>If you'd rather keep it symmetric with the defense path, the alternative is to patch the 0x4A1280 attacker-preference comparator the same way prefer_first_defender is patched and drop the override entirely. Let me know which you prefer.

It's fine as is.

We're almost done now. I ran this branch through a bunch of in-game tests and it passed all of them, only unfortunately I did find one major issue while experimenting with armies. Other than that, there are a few minor issues with the code that I saw reading through the recent commits. First, the major issue: The setup is that you have an army selected with two enemy units on different adjacent tiles. I don't know if it matters but in my test the units had different preferred attackers from the army. If you enter go-to mode and mouse over the enemies, the unit shown on the army changes as expected. However, while you're mousing over one unit, if you launch an attack against the other one by pressing an arrow key, the game will freeze playing the army guy's animation.

The minor issues I found are:

- On line 30022 (in `counter_attacker_for_defender_selection`), the arguments to `clamp` are in the wrong order.
- `tiles_adjacent_for_attack` isn't necessary; there's already a function for that called `are_tiles_adjacent`.
- `face_unit_toward` doesn't work over map edges and also isn't really necessary since the base game's `Map::compute_neighbor_index` function does basically the same thing. Note that the neighbor indices 0 through 8 correspond with the directions enum.
- I think the names of the `unit_group` and `counter_rule` config options should be made plural so they're `unit_groups` and `counter_rules`. That way it's clearer that they accept lists. I used Codex to write the in-game tests I mentioned and this actually confused the AI. It wrote a config with multiple `counter_rule` entries for multiple rules instead of gathering them into one. Players may get confused too.
- This is not necessarily a bug but an observation: The logic to define what's a river crossing is different for the counter rules compared to the base game. In the base game, to apply the crossing penalty, it checks only the river code on the defender's tile, whereas for the counter rules an attack is considered to cross a river if either the attacker's or defender's tiles have a river code bit set. This may never matter in practice if the map generator and editor always set both river codes, but still there's no reason for the logic to be different.
- Another observation: The game shows the counter-aware top defender against bombardment when a unit is selected and the bombard button is active, but just because that button is active doesn't mean the unit is an artillery, for example many ships can attack or bombard. Maybe it's better to show the bombard defender only when the button is active and (attack strength == 0 or type is air unit)?

maxpetul:
One last thing. This comment in the config:
```
;   Multiple rules can be placed in one counter_rules list, separated by commas:
;   counter_rules = [ranged vs melee self-atk 125], ["Knight" vs melee in-city enemy-def 150]
```
is not correct. There can only be one set of square brackets and the comma must be inside like so: `[ranged vs melee self-atk 125, "Knight" vs melee in-city enemy-def 150]`.

maxpetul:
Great, this all looks good to me. Thanks for the contribution.

maxpetul:
FYI: I've merged the `unit_groups` and `unit_limit_groups` config options into a new `unit_type_tags` config option (relevant commit: c40b89b). The format is the same so you can copy over the contents from the old options if you're using them. I could have recognized the old option names as synonyms for `unit_type_tags` but chose not to because those old options weren't present in any official release so I didn't feel it was necessary to keep backwards compatiblity.

The idea behind the type tags is to have a single place where you can define categories of unit types. Beyond combining the group options, I extended the tags so they can be put anywhere that the config previously expected a unit type. Also, I made it so that tags can be defined over ranges of units connected by upgrades, and so that they can be defined recursively. For example:
```
unit_type_tags = [HorsebackUnits: Horseman   "Mounted Warrior"   Knight to Rider   Cavalry to Cossak   "Ancient Cavalry"   Conquistador,
                  HorseUnits: HorsebackUnits Chariot "War Chariot" "Three-Man Chariot"]
```
This defines a "HorsebackUnits" tag for all base game units that ride on horseback and a "HorseUnits" tag for all units that include a horse.

This was the last thing I wanted to do before the next release, so R28 Preview 1 is coming very soon.

PR #32, opened 2026-05-02
Fix workers on coast waking up by setting unit state; Loosen AI worke…

instafluff0:
Howdy - this PR

1. Fixes the issue of workers on coastal tiles with a path set waking up each turn. It was silly - I had forgotten to set `this->Body.UnitState = prev_state;` after running the base `Unit_move_to_adjacent_tile`.
2. Loosens restrictions on AI workers for building ports via coastal movement checks. Previously AI worker movement was checked by ensuring the workers were already assigned to build ports. It's difficult to know for sure this was the issue, but loosening that to only check if they are on their own territory *appears* to lead to earlier port building. In my tests I found civs on vanilla speed, non-accelerated production starting around 400 BC. Other civs don't do so until much later. Given the number of parameters and subjectivity in measurement, it's hard to know whether this is actually an issue or not still. 

The reason for the AI coastal worker checks in the first place is just how awkward (almost like cheating though not, as the player can do it) it looks for many AI workers to be moving freely across coastal waters on shortest paths somewhere. I found limiting it to just their territory looks much more natural and less "cheat-y".

Also please note I was getting a compile error with `do_not_place_barb_huts_or_camps_on_impassable_tiles` not being in C3X.h. I don't know if that's an oversight or merge issue, etc. but to get everything to compile on my end I added it to the end. You may not want it declared there; just let me know.

Last - I'm still working on the issue with `naval_units_use_port_districts_not_cities` causing AI units not to be loaded on transports. I expect to finalize and have a PR for that this weekend but will keep you in the loop if not.


maxpetul:
The error with `do_not_place_barb_huts_or_camps_on_impassable_tiles` is because originally that option only covered goody huts and not barb camps, and was named `do_not_place_barbarian_huts_on_impassable_tiles`. When I extended it to cover camps I must have forgotten to update C3X.h (though I'm sure I tested it...). I'll fix that on master now.

Other than that, this PR looks fine. At the risk of complicating things, perhaps it would make sense if workers were allowed to enter coast tiles only when moving from a land or bridge tile? So they could enter coast to do work but after that their only option would be to return to land. That way they wouldn't be able to take shortcuts over water or walk between continents, etc.

instafluff0:
> The error with do_not_place_barb_huts_or_camps_on_impassable_tiles is because originally that option only covered goody huts and not barb camps

Ok sounds good.

---

> At the risk of complicating things, perhaps it would make sense if workers were allowed to enter coast tiles only when moving from a land or bridge tile? So they could enter coast to do work but after that their only option would be to return to land.

This makes sense to me and seems pretty reasonable. I think it also depends though on whether we eventually implement some sort of switch to a worker boat FLC for workers on coastal tiles. That's something a number of folks have suggested in the forums and DMs to me. Civinator for example suggested having C3X switch the worker to the Walk FLC, if defined in the animation INI file (he asserted that Walk is otherwise unused in the game, which I can't confirm).

I haven't committed to doing this yet, as it seems like a nice-to-have, and I don't know how much work it would be. If workers on coasts looked like boats, it would be perhaps unexpected if they couldn't move to another coastal tile. On the other hand, if they stayed in the vanilla FLC look, it makes a lot of sense.

What do you think?

Edit: just to be clear, I'm not thinking to implement the water FLC change for R27; if I were to do it at all, it would be down the road.

maxpetul:
Playing the walk animation might be difficult because you'd have to load it first. There are many animations listed in the INI files that the engine does not use at all. See `animation_names`, a list of 19 strings at 0x680758. Those are the only animations that the game will load when it loads a unit INI file and the walk animation is not one of them. In fact "WALK" does not appear anywhere in the program so Civinator is surely right that it's not used.

Besides the animation issue, the other reason I'd be reluctant to allow workers to walk wherever they please over coast is that it upsets the game balance a bit. Like, would you ever build a curragh if you could send a worker out to explore along the coast? The AI can be prevented from doing that but the temptation would still be there for human players.

instafluff0:
> Playing the walk animation might be difficult because you'd have to load it first. There are many animations listed in the INI files that the engine does not use at all. See...

Interesting, got it. Well I think that answers that.

> Besides the animation issue, the other reason I'd be reluctant to allow workers to walk wherever they please over coast is that it upsets the game balance a bit. Like, would you ever build a curragh if you could send a worker out to explore along the coast?

Agreed, that's true too. I'll try to restrict worker coastal movement as you suggest. And speaking of curraghs, I belatedly realized that setting the default port tech requirement as Map Making means that effectively no one would ever build them; naval units would start with galleys. The upcoming commit(s) will have the default Port tech as Alphabet, so at least curraghs could be built.

Also, I think I've finally sorted out the issue with naval transports and ports (here's a couple debug screenshots):

Screenshot 2026-05-05 at 4 46 05 PM

Screenshot 2026-05-05 at 4 34 58 PM

I need to do a bit of cleanup and will post an update here when ready again for review.

instafluff0:
Ok, ready for review (and merging, if looks good). This PR

1. Fixes issue with workers on coastal path waking up each turn (discussed above).
2. Fixes the issue of settlers and offensive/defensive units not loading in transports. Units get redirected to look for port tiles for a given city, and if `patch_Unit_load` is called, either moves the unit toward the port or loads them directly. With this change I've observed AI cross-continent settlement and invasions working fine. Note that part of this includes changing the port defense logic, such that if a naval unit has any land units loaded, it uses vanilla movement logic and doesn't try to defend the port. (if we didn't do so, I found galleys with units simply sitting in ports, as they were the sole defenders)
3. Restricts worker movement to coastal tiles only if originating from land or a bridge.
4. Prevents automatically adding walls to cities on same continent as Great Wall wonder if `disable_great_wall_city_defense_bonus` is true, in addition to preventing the defensive bonus itself. This was my original intent when I implemented the feature (ie, walls only around the civ territory, not cities). Note that it still allows individual cities to build walls just fine.
5. Sets the default Port tech prereq as Alphabet (rather than Map Making), so Curraghs can still be built.
6. (minor) only visually connect great wall districts if same territory owner (this was something haluu mentioned)

I've tested all features and have saves for each, which I'm happy to share. Based on our convo I don't intend to do anything with worker boats animations (in this PR or in the future, right now).

After this is merged, I'll prepare PRs for the R28 seasons & animations features.

maxpetul:
This all looks good to me. I'll merge it sometime soon-ish. I want to merge PR #31 first, which shouldn't take much longer.

maxpetul:
After experimenting with it, I see that other PR might need some time so I'll just merge this now.

PR #33, opened 2026-05-08
Great wall multi civ

drhaluu82-dev:
A way for all civs to build Great Walls

* All Civs can auto-build walls (once per civ)
* End user (modder) should set the building that allows the Great Walls auto-build to a Small Wonder so that all Civs can build, in C3X_config.ini setting "great_wall_auto_build_wonder_name".  Building perfumes can be used for this small wonder to increase/decrease the likelihood that AI will auto-build walls.  (if great_wall_auto_build_wonder_name is set to a Great Wonder, then only one civ can auto-build walls like before).  
* Ensures that AI Wall auto-builds cannot be done on tiles adjacent to opposition walls.  This keeps walls from overcluttering the map, and also keeps AI economy up (since by default walls do not provide any yields, so too many walls can hurt AI economy in-game).  Also aesthetically more pleasing, as walls seem to autoconnect with enemy walls when adjacent, and it looks very strange.
* Also adds optimization for the new Great Walls feature.  When I first tested, slowdowns were happening in the late game.  So I added optimization to protect against that.  Tested and works great in game, without any noticable lag/slowdowns.  
*  You will see project #'s like ToC-3, ToC-5, and ToC-24 in the comments, Those are just different points at which improvements to the walls were added.
* Screenshots attached.  
Screen5-Walls
Screen6-walls


maxpetul:
Hello, haluu. Thanks for the contribution. This looks good overall except the PR is missing the changes to C3X.h so it doesn't compile as `great_wall_auto_build_done_civs` is not defined in `injected_state`. Also, an extremely minor issue: the comment "// If the worker has arrived" was duplicated.

I have a question. Maybe I'm missing something obvious, but what's the point of the "re-entrancy guard" on `auto_build_great_wall_districts_for_civ`? The comment says it's there in case "the function is triggered for a civ while it is already mid-run for another", but how is that possible? Maybe in simultaneous multiplayer? But even then you wouldn't want to lock players out from building the wall when their attempt to build it coincides with another player's.

drhaluu82-dev:
@maxpetul thanks for Reviewing!!  made updates both the first 2 issues you pointed out, you should see them in this GitHub Pull Request now.  

For the re-entrancy guard, I looked at this a bit more, and you are right! not necessary.  Also, didn't factor in multiplayer when coding this (combination of vibe coding and my own edits), so will do so moving forward. 

I removed that re-entrancy guard, tested in-game, and multiple civs where still able to build their own Great Walls.  

Ready again for your code review!  thanks!  

Just a note to self so I can remember, this was the line I removed (previously in line 24399): 
```
	// ToC-3: Re-entrancy guard — if a previous call is still executing (GWABS_RUNNING),
	// do not start a second pass.  This prevents double-processing if the function is
	// triggered for a civ while it is already mid-run for another.
	if (is->great_wall_auto_build == GWABS_RUNNING)
		return;
```

drhaluu82-dev (commented):
updated header file to include great_wall_auto_build_done_civs

drhaluu82-dev (commented):
removed re-entrancy guard

maxpetul:
The `great_wall_auto_build` variable is no longer needed at all and can be deleted. I went ahead and took care of that. Now, if that variable is loaded from a save and it's set to 2 (GWABS_DONE), that's handled by setting `great_wall_auto_build_done_civs` to 0xFFF... like for the other obsolete variable `great_wall_auto_build_is_done`. Otherwise it's ignored. Does this look right to you?

drhaluu82-dev:
@maxpetul 
Yes, that looks correct to me. Setting great_wall_auto_build_done_civs = 0xFFFFFFFF when the old GWABS_DONE state (value 2) is found in the save is exactly the right behavior — it prevents any civ from re-running the auto-build on a map that was already walled before the save. The parallel with the older great_wall_auto_build_is_done chunk makes it easy to follow too. No concerns on my end.

One unrelated issue... just pulled the most recent version of the changes here and compiled on my local machine, and tried to load a previously saved game, but this got this soft (able to be bypassed) error below.  Also got the same error when loading up a new game.  

error1
I think somewhere along the way, the default.c3x_config.ini file got mispelled for this variable.  Since it was correct on C3X Preview 2 public release, but in the main fork for the current C3X in GitHUB, i'm seeing that default.c3x_config.ini is mispelled.  Should be "impassible" with an "i" in the config file to match the code files, not "impassable" with an "a" I believe.

Screen2error

instafluff0:
I think @maxpetul corrected my incorrect spelling in this case.

maxpetul:
Great, I'll go ahead and merge this PR. Thanks again for the contribution.

About that error: Actually "impassable" was originally misspelled as -ible and I recently changed it to -able. "Impassable" and "impassible" are both real words. The first means what you'd expect, "not able to be passed", while the second means "unemotional" more or less, similar to "impassive". If you've updated everything it should work, it looks like you've got the newer config files but not the newer code files somehow.

instafluff0:
> "Impassable" and "impassible" are both real words. The first means what you'd expect, "not able to be passed", while the second means "unemotional" more or less, similar to "impassive"...

That's hilarious, I had no idea. My confidence from winning the 6th grade spelling bee has been shaken to its core.

maxpetul:
>That's hilarious, I had no idea. My confidence from winning the 6th grade spelling bee has been shaken to its core.

I know, right? Who came up with this language anyway?

PR #34, opened 2026-05-08
R28 tile improvements

instafluff0:
This is my PR one of three for R28. Those are:

1. r28-tile-improvements (this)
2. r28-seasonal-cycle
3. r28-animations

I'd suggest merging them in that order, as (3) was branched from (2) and depends on the seasonal-cycle code.

This PR

1. Allows for districts as genuine custom tile improvements used by only one city at a time, using the config:

  ```
  type                   = tile-improvement
ai_build_strategy      = tile-improvement
  ```
  
  tile-improvements
  
2. Allows workers to be consumed and despawned after completing a district, like colonies.

   ```
   consumes_worker = 1
   ```
3. Allows the default resource on a tile to be suppressed if a district is built over it

   ```
   subsumes_tile_resource = 1
   ```


injected_code.c
```diff
@@ -6399,9 +6408,27 @@ district_is_complete(Tile * tile, int district_id)
...
+			Unit * worker_to_consume = NULL;
+			FOR_UNITS_ON (uti, tile) {
+				Unit * unit = uti.unit;
+				if ((unit != NULL) && is_worker (unit)) {
+					worker_to_consume = unit;
+					break;
+				}
+			}
+			if (worker_to_consume != NULL) {
+				if ((p_main_screen_form != NULL) &&
```
maxpetul: It's safe to assume that `p_main_screen_form` will never be NULL. It's not even a variable, it's a preprocessor macro defined by the patcher using the address and type set in civ_prog_objects.

injected_code.c
```diff
@@ -6399,9 +6408,27 @@ district_is_complete(Tile * tile, int district_id)
...
+			FOR_UNITS_ON (uti, tile) {
+				Unit * unit = uti.unit;
+				if ((unit != NULL) && is_worker (unit)) {
+					worker_to_consume = unit;
+					break;
+				}
+			}
+			if (worker_to_consume != NULL) {
+				if ((p_main_screen_form != NULL) && 
+					patch_Leader_is_tile_visible (&leaders[p_main_screen_form->Player_CivID], __, worker_to_consume->Body.X, worker_to_consume->Body.Y)) {
```
maxpetul: Surely the worker should get despawned even when no one's looking?

injected_code.c
```diff
@@ -9547,13 +9614,22 @@ handle_district_definition_key (struct parsed_district_definition * def,
...
 			def->auto_add_railroad = (ival != 0);
 			def->has_auto_add_railroad = true;
-		} else
-			add_key_parse_error (parse_errors, line_number, key, value, "(expected integer)");
-
-	// Match misspelling of "impassable" too so old configs still load without error
-	} else if (slice_matches_str (key, "impassable") || slice_matches_str (key, "impassible")) {
-		struct string_slice val_slice = *value;
-		int ival;
+		} else
```
maxpetul: Notice how the lines around this insertion are considered to have been changed even though their contents is the same? That indicates the line endings have been changed. Please make sure that all lines in injected_code.c have dos-style endings. This seems to be a recurring issue with Codex and maybe other AI code editors that the edits they make have unix-style endings regardless of what's standard in the file they're editing. I had this problem before so I had Codex write fix_line_endings.py, which ensures that all lines in all files of the project have the expected ending style. That may be useful for you.

injected_code.c
```diff
@@ -28684,6 +28877,40 @@ patch_City_calc_tile_yield_while_gathering (City * this, int edx, YieldKind kind
 		}
 	}
 
+	if (is->current_config.enable_districts) {
```
maxpetul: What is the purpose of this block of code? For food and shields it's redundant because the base `City::calc_tile_yield_at` returns the values from `Map::calc_food_yield_at` and `Map::calc_shield_yield_at`, and those `Map` functions are already patched to handle improvement districts and generated resources. This block is necessary to calculate commerce yield correctly, but why place that here instead of in a patch for `Map::calc_commerce_yield_at`?

maxpetul (changes requested):

injected_code.c
instafluff0: ~~I know what you mean, but I'm not sure this is a case of that. I just ran it and I don't see anything with line endings off in injected_code.c. Am I misunderstanding the script?~~

Edit: scratch that, I think I had my local branch in a funny state due to a misbegotten git reset, trying again...

Edit 2: Ok, should be good now. I'll run this for the other branches as well.

injected_code.c
instafluff0: Ah, sorry this was a dumb half-baked leftover from my experiments to show worker animations when they despawn, rather than simply disappearing. Because the camera was always on the tile when I finalized the code I hadn't realized I'd even left it.

Thanks for catching that - removed now.

instafluff0 (commented):
Please see comments below.

injected_code.c
instafluff0: Right. Short answer: I got a crash every time I tried to patch `Map::calc_commerce_yield_at`.

If I remove this block and add:

`inlead,    0x5d7AD0,    0x0,        0x0,        "Map_calc_commerce_yield_at",			"int (__fastcall *) (Map * this, int edx, int tile_x, int tile_y)"`

```
int __fastcall
patch_Map_calc_commerce_yield_at (Map * this, int edx, int tile_x, int tile_y)
{
	if (! is->current_config.enable_districts)
		return Map_calc_commerce_yield_at (this, __, tile_x, tile_y);

	Tile * tile = tile_at (tile_x, tile_y);
	if ((tile != NULL) && (tile != p_null_tile)) {
		struct district_instance * inst = get_district_instance (tile);
		if (inst != NULL &&
		    district_is_complete (tile, inst->district_id)) {
			if (! district_uses_tile_improvement_rules (inst->district_id))
				return 0;
			City * yield_city = get_city_ptr (tile->Body.CityAreaID);
			if (! district_tile_bonus_applies_to_city (tile, inst->district_id, yield_city))
				return 0;
			struct district_config * cfg = &is->district_configs[inst->district_id];
			int commerce_bonus = 0;
			get_effective_district_yields (inst, cfg, NULL, NULL, &commerce_bonus, NULL, NULL, NULL);
			if ((cfg->generated_resource_id >= 0) &&
			    (cfg->generated_resource_flags & MF_YIELDS) &&
			    district_generates_resource_for_civ (tile, inst, cfg, yield_city->Body.CivID)) {
				Resource_Type * res = &p_bic_data->ResourceTypes[cfg->generated_resource_id];
				commerce_bonus += res->Commerce;
			}
			return commerce_bonus;
		}
	}

	return Map_calc_commerce_yield_at (this, __, tile_x, tile_y);
}
```

I get a crash upon founding a city or trying to open the city screen. I could not figure out why and probably should have tried to determine if the arguments we had for that function in Ghidra were correct, but did not, so this seemed a reasonable workaround. But I hear you and agree it is duplicative. I think my development of this branch also pre-dated our earlier discussion of how to figure this stuff out using Ghidra. 

Let me take another stab at it.

instafluff0: Update: I believe I've got things running now and corrected. Ghidra had `RET 0x18` for `Map::calc_commerce_yield_at`, so 6 args, which I'm fairly certain were exactly the same as `Map::calc_food_yield_at`.

I updated civ_prog_objects accordingly, removed the block in  `patch_City_calc_tile_yield_while_gathering`, and added `patch_Map_calc_commerce_yield_at`.

instafluff0 (commented):
Please see comments inline regarding `Map::calc_commerce_yield_at`.

maxpetul:
Great, that all looks good now. There's one little cosmetic thing I just noticed: the commerce gathered from improvement-type districts is not outlined in purple in the "COMMERCE" area of the lower frame of the city screen. That's unlike shields from improv-type districts and commerce from normal-type ones, which are both outlined in purple.

And that reminds me of another cosmetic thing I didn't mention before: when you hover over an improv-type district that's being worked on the city screen, the game doesn't highlight the tile yields like it normally does. I figured you were aware of that and decided it wasn't worth bothering with, but I'm pointing it out just in case. I'm not insisting you change it, if you really think it's not worth the effort, I'm fine merging this PR as is.

instafluff0:
> There's one little cosmetic thing I just noticed: the commerce gathered from improvement-type districts is not outlined in purple in the "COMMERCE" area of the lower frame of the city screen. That's unlike shields from improv-type districts and commerce from normal-type ones, which are both outlined in purple.

Yeah, that's true. I was aware of it but am wary (lazy) to revisit more code to try to get the positioning *just right* to either intercept the current code and draw differently or write a post-processing function to draw the purple-lined ones on top of it. It was just a huge amount of work to do that the first time around, and I'm hoping no one sees it as a big deal to not have those around commerce.

---

> And that reminds me of another cosmetic thing I didn't mention before: when you hover over an improv-type district that's being worked on the city screen, the game doesn't highlight the tile yields like it normally does.

Yeah I noticed that too. In fact, my initial thought was if these are to behave like real tile improvements, they shouldn't even have the purple outlines, in which case I could just call the vanilla code (which would also handle the highlighting). I realized that wouldn't work though as custom tile improvements would still allow for science, culture, etc. benefits, which wouldn't get drawn if I delegated to the vanilla code. So I thought it mostly reasonable to just leave as-is and accept they wouldn't do the typical tile highlights.

If you're cool with merging as-is I'd vote we do that.

maxpetul:
Good enough for me.

PR #35, opened 2026-05-08
R28 seasonal cycle

instafluff0:
PR two of three for R28. Adds the seasonal cycle, same general approach as day-night cycle.

One major thing to note: this PR changes the default path assumption for districts & day-night art from

`Art/DayNight/1200`
`Art/Districts/1200`

to

`Art/DayNight/Summer/1200`
`Art/Districts/Summer/1200`

...with other seasons similarly structured (in place of Summer in this example) and day-night hours similarly nested under them.


injected_code.c
```diff
@@ -17311,223 +17363,432 @@ bool load_day_night_hour_images(struct day_night_cycle_img_set *this, const char
...
+				is->current_seasonal_cycle = output;
+				QueryPerformanceCounter (&is->last_seasonal_cycle_update_time);
+			}
+
+			LARGE_INTEGER perf_freq, time_now;
+			QueryPerformanceFrequency (&perf_freq);
+			QueryPerformanceCounter (&time_now);
+
+			double elapsed_seconds =
+				(double)(time_now.QuadPart - is->last_seasonal_cycle_update_time.QuadPart) /
```
maxpetul, line 17736: `last_seasonal_cycle_update_time` gets used here without having been initialized when the game was loaded from a save because the loading code reads `current_seasonal_cycle` then sets `seasonal_cycle_unstarted` to false without ever setting `update_time`.

maxpetul (changes requested):

injected_code.c
instafluff0: Thanks - updated now to initialize when loaded from a save

```c
} else if (match_save_chunk_name (&cursor, "current_seasonal_cycle")) {
    is->current_seasonal_cycle = clamp (CS_SUMMER, CS_SPRING, *((int *)cursor)++);
    QueryPerformanceCounter (&is->last_seasonal_cycle_update_time);
    is->seasonal_cycle_unstarted = false;
    if ((is->day_night_cycle_img_state == IS_OK) && ! is->day_night_cycle_img_proxies_indexed)
        build_sprite_proxies (&p_bic_data->Map.Renderer);
}
```

Also, in case you want to try the art out, link for the files is at https://drive.google.com/file/d/1tT2Gz8L_WKvzwr_xp2lusjz9cNCph9s6/view?usp=drive_link

maxpetul:
I tried kicking the tires on the various seasonal cycle modes. The good news is that four of them work, the bad news is that on-day-night-hour is broken quite badly. When I set that and set `day_night_cycle_mode` to every-turn, the game gets stuck while loading with 30% on the bar. It pauses for a while there then shows a popup message, "Unable to allocate draw-buffer; terminating program.", and after that becomes unresponsive and must be force quit.

Interestingly, I can get the game to load if I first load the test scenario with `day_night_cycle_mode` off then quit to main menu, set it to every-turn in the scenario config, and load the scenario again. However that produces corrupt graphics on the map, this is after moving the camera up and down:
day_night_seasons_graphical_corruption
Not sure if this is a separate issue.

I could send you my test scenario if you'd like but it's very simple. I dropped those files from your Google Drive link in the scenario's Art folder and my scenario config is:
```
day_night_cycle_mode = every-turn
seasonal_cycle_mode = on-day-night-hour
transition_season_on_day_night_hour = 3
```

instafluff0:
That is really wacky, thanks for letting me know. I haven't tried the 2nd thing you mention (start with it off, then change config and get corrupt graphics), but I haven't yet been able to reproduce the "Unable to allocate draw-buffer; terminating program" error. 

Loading all of that art definitely takes a while, but eventually it works for me:


And I can get through to the actual transitions just fine (this is actually the mode I intend to use when I play with, so I've been testing with it):


image

How would you suggest approaching this? You're much more knowledgeable about memory management than me in this case, so if you'd suggest I restructure any of this I'm happy to. Just not sure where to begin.

**Edit**: I noticed that there is a background process that is using a ton more memory as well. I've actually known about this since the day/night cycle work, as I found after closing the game, if the day/night cycle is enabled, one can't restart it if this background process is still running. And with the seasonal cycle also enabled, it takes even longer to close, which is annoying. I have no idea what it's doing, nor why garbage collection (if that's what's happening?) takes so long.

Anyway I'm not sure if this is at all related, but thought I'd point it out.

image

instafluff0:
One more update: as a quick experiment (I haven't committed or pushed) I had Codex rewrite all cycle code to lazy load and cache instead, on demand. I haven't extensively reviewed it.

I've only briefly tested, but performance is much better than I expected in-game, and obviously main menu loading is lightning fast and feels like a normal game. Every PCX file for a given season/hour is still loaded (regardless of whether it is proxied and shown or not), but on my machine I see only a slight delay between turns, if any, at the beginning of the game. Art is still cached though, so over the course of 4*24 turns, the memory footprint is just as large as it is now.

If you think this is worth pursuing, let me know and I can review the code and commit it.

Alternatively, if we like this direction but don't want the memory footprint to be so big, I suppose we could only keep in memory the art needed for that particular season/hour, maybe only using this strategy if the day/night cycle & seasonal cycle are used together? Just thinking out loud.

lazy

maxpetul:
Yeah, the issue is simply that it's running out of memory. I was testing with RUN.bat so the LAA bit wasn't set and then the game was crashing when it reached the 2 GB limit. It loads eventually and works fine if I install first before testing. However the graphical corruption issue is still there after installing. I noticed too that it loads very quickly in that case, as if it's never loading the day/night and season graphics after the feature is turned on and is instead reusing something that was never initialized.
>Alternatively, if we like this direction but don't want the memory footprint to be so big, I suppose we could only keep in memory the art needed for that particular season/hour, maybe only using this strategy if the day/night cycle & seasonal cycle are used together? Just thinking out loud.

That would be my preferred approach. The only reason not to do that would be if reloading the graphics takes too long, but I don't expect it to and it looks like it doesn't based on your brief test. Even if there's a noticeable pause of the few seconds, player's won't mind and probably won't even notice if it's folded in with the overall wait between turns. Although unloading graphics might be a problem if that's what's causing the game to take so long to shut down. I don't know what causes that and actually have never noticed it before.

Losing ~2 GB of memory for the season and day/night graphics is too much in my opinion. Even just ~0.5 GB for the day/night graphics alone is kind of a lot. Even without losing that memory it's possible for the game to use up the entire >3 GB expanded 32-bit address space, and it happens in real games because there have been people complaining about that in the mod thread on CivFanatics.

By the way, is it convenient for you if this memory issue is handled in a separate PR? If so, I'd rather do it that way. After the graphical corruption issue is fixed, I'd consider the seasonal cycle feature complete and tested, and this PR ready to merge.

instafluff0:
> However the graphical corruption issue is still there after installing. I noticed too that it loads very quickly in that case, as if it's never loading the day/night and season graphics after the feature is turned on and is instead reusing something that was never initialized.

Yeah, I was able to reproduce the issue and did some debugging and logging and believe I've figured it out. And the more I think about it, the more I agree that we should definitely just load the season+hour art needed - under any configuration, if either or both are on - for that given turn. That would just simplify so much of the bookkeeping and headaches here for handling users exiting and reloading/starting over in the game.

> By the way, is it convenient for you if this memory issue is handled in a separate PR? If so, I'd rather do it that way. After the graphical corruption issue is fixed, I'd consider the seasonal cycle feature complete and tested, and this PR ready to merge.

On that note, if you'd like that to be a separate PR, that's fine with me. I think making that architectural change also incidentally solves the graphical corruption problem as well though, so, I dunno, it seems unnecessary to me for us to merge this one in if we're shortly thereafter just going to simplify the architecture in a separate PR. But it's up to you and I'm fine either way.

Anyway from what I can tell here is the issue:

First load of scenario:
1. Day-night is `off`, seasonal is `on-day-night-hour`
2. 4 sets are expected (Fall/1200, etc.), all indexed correctly in 4 of 96 available slots.
3. `is->day_night_cycle_img_state` is set to `IS_OK`

Second load:
1. Day-night is `every-turn`, seasonal is `on-day-night-hour`
2. At runtime we'll expect to swap in 4 x 24 `day_night_cycle_img_set`s
3. But `is->day_night_cycle_img_state` is `IS_OK`, so `init_day_night_and_seasonal_images()` doesn't reload the remaining 92 image sets
4. `build_sprite_proxies` runs assuming 96 proxied sets, when only 4 are actually loaded

So it builds proxies into uninitialized/zeroed sprite slots for the 92 hour/season combinations that were never loaded. When the game later draws at a non-noon hour, it substitutes bogus sprites, which matches the corrupt overlay/trail behavior.

One way to look at this is that semantically `is->day_night_cycle_img_state` should not be shared between seasons and day-night. But that's also yet more bookkeeping which adds to the complexity.

I just made a commit (https://github.com/maxpetul/C3X/pull/35/changes/1fe57ea18e3771140259151db0040079c6ee6e1f) that sets seasonal mode to off if it is set to depend on the day-night cycle, which in turn is off. I tested and this issue with the corrupt graphics doesn't happen to me anymore. That's also setting aside though the question of whether - in the current "cache everything" approach - we want to even actively support users changing the cycle config settings in mid-game.

---

As I said though, I agree and would like to just simplify all of this and grab & store only the art we need for a given season + hour. That would make this particular issue (and commit above) moot, solve the memory usage issues, and also reduce the bookkeeping and state checks we need to do.

instafluff0:
Update: in the course of testing, I just came upon this ... https://github.com/maxpetul/C3X/issues/45

maxpetul:
>Anyway from what I can tell here is the issue: ...

That makes sense. My approach would have been to deinit and reinit the images whenever a game is loaded because you don't know what's changed in the meantime. Not only might the player have changed the config, but they also might have changed the art files. But keeping only one set of images and loading them just before they're going to be displayed makes this a moot point.

I'll go ahead and merge this because the seasonal cycle per se is done. I'm looking forward to another one with the changes to image loading.

PR #40, opened 2026-05-13
Add grouped unit limits feature

drhaluu82-dev:
* Allows for end user (modder) to designate Unit Limits based on Unit Groups.  Can add certain unit types to a unit group, and the Max for the entire group can be determined using the existing unit limit system.  
* Also submitted the config file with new field as an example: 

**; Here it's possible to limit how many units of each type players may build. Example:
;   ["Royal Guard": 3, Champion: 1 per-city, "Heavy Tank": 3 cities-per, "Heavy Infantry": 5 + 2 per-city]
; The limits may be constant values or vary depending on city counts. In the example above, players will each be limited to 3 Royal Guards. They may
; build one Champion for each city they own. They may build one Heavy Tank for every 3 cities they own. It is also possible to combine terms with plus
; signs, as for Heavy Infantry.
; The unit limits apply to unit production by players, not all forms of unit creation. Specifically, they apply to city production, upgrading, and
; auto-production from wonders. They do not apply to all other ways units may be created including by being captured, enslaved, spawned from a razed
; city, spawned for barbarians, pre-placed in a scenario, spawned for the AI based on difficulty level, etc.
; Can also add unit_limit_groups here too in order to limit entire unit group to x amount.  See "unit_limit_groups".
unit_limits = ["Infantry Units": 2]

; Unit Limit groups -- All units in the unit limit group count towards the limit.  If a unit name does not exist in unit_limits,
; then unit_limits checks unit_limit_groups for group names as well.  Groups are separated by commas, and units within are group
; are separated by spaces.
; For instance:
;   unit_limit_groups = ["Infantry Units": "Castle Guard" "Guardian",
;                     "Tanks": "Light Tank" "Heavy Tank" "Panzer"]
; then in unit_limits setting (above), set a limit for the unit group:
;   unit_limits = ["Infantry Units": 4, "Tanks": 3 cities-per]
unit_limit_groups = ["Infantry Units": Warrior Spearman]**

* This has been tested in-game (vanilla) and works.  In this example, Warrior/Spearman are in the same unit limit group, so only 2 of them combined can be built.  When 2 are built, can't build a third until we lose a Warrior/Spearman.  
test1

* You will see ToC-26 and ToC-27 -- These are project numbers that can help you find the code additions/replacements. 
* ToC-26 : Build Unit Group Limit system
* ToC-27 : Add functionality to ensure that upgrades are still allowed if max limit has already been reached, and a unit within the group wants to upgrade to another unit type within the same unit limit group.

### DIFF
--- a/C3X.h
+++ b/C3X.h
@@ -78,6 +78,17 @@ struct unit_type_limit {
 	int cities_per;
 };
 
+/* ToC-26: Represents a named group of unit types that share a combined build limit.
+   Stored by pointer in unit_limit_groups (keyed by group name) and in unit_type_to_group
+   (keyed by unit_type_id).  The group struct is owned by unit_limit_groups; unit_type_to_group
+   holds non-owning pointers into it. */
+struct unit_limit_group {
+    int * unit_type_ids;        // Array of unit type IDs belonging to this group
+    int   count;                // Number of IDs in the array
+    struct unit_type_limit limit; // The shared limit value for this group
+    bool  has_limit;            // True once a limit has been assigned via unit_limits
+}; // END ToC-26
+
 struct work_area_improvement {
 	short improv_id;
 	int work_area_radius_limit;
@@ -320,6 +331,10 @@ struct c3x_config {
 	struct leader_era_alias_list * leader_era_alias_lists;
 	int count_leader_era_alias_lists;
 	struct table unit_limits; // Maps unit type names (strings) to pointers to limit objects (struct unit_type_limit *)
+	// ToC-26: Group-based unit limits. unit_limit_groups maps group name strings to struct unit_limit_group*.
+	// unit_type_to_group maps unit_type_id (int) to struct unit_limit_group* for O(1) runtime lookup.
+	struct table unit_limit_groups;
+	struct table unit_type_to_group; // END ToC-26
 	bool allow_upgrades_in_any_city;
 	bool do_not_generate_volcanos;
 	bool do_not_pollute_impassable_tiles;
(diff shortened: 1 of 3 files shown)


drhaluu82-dev:
noticed this typo in config file, will fix. 
image

maxpetul:
This all looks alright. I'd rather have one way of categorizing unit types instead of having separate unit limit groups and unit counter groups (from PR #31). It would be nice then if the mod config could take a unit group anywhere that it already takes a unit type like, for example, for perfuming. However, if we have generic unit groups, we'll also want to allow unit types to be in multiple groups, f.e. a knight ought to able to be in a "mounted" group as well as a "melee" group. Then those groups could have different limits, which would add a layer of complication to the logic here. Anyway, those are all matters for the future. This PR is fine, I'll go ahead and merge it.

drhaluu82-dev:
@maxpetul great callout on a future general grouping system!  will keep that in mind for sure!  I love the idea.  

and thanks!

maxpetul:
FYI: I've merged the `unit_groups` and `unit_limit_groups` config options into a new `unit_type_tags` config option (relevant commit: c40b89b). The format is the same so you can copy over the contents from the old options if you're using them. I could have recognized the old option names as synonyms for `unit_type_tags` but chose not to because those old options weren't present in any official release so I didn't feel it was necessary to keep backwards compatiblity.

The idea behind the type tags is to have a single place where you can define categories of unit types. Beyond combining the group options, I extended the tags so they can be put anywhere that the config previously expected a unit type. Also, I made it so that tags can be defined over ranges of units connected by upgrades, and so that they can be defined recursively. For example:
```
unit_type_tags = [HorsebackUnits: Horseman   "Mounted Warrior"   Knight to Rider   Cavalry to Cossak   "Ancient Cavalry"   Conquistador,
                  HorseUnits: HorsebackUnits Chariot "War Chariot" "Three-Man Chariot"]
```
This defines a "HorsebackUnits" tag for all base game units that ride on horseback and a "HorseUnits" tag for all units that include a horse.

This was the last thing I wanted to do before the next release, so R28 Preview 1 is coming very soon.

drhaluu82-dev:
@maxpetul thanks so much Flintlock!  Wow I love it!
Agreed you don't need to create a synonym on the unreleased config variable.  I think for my version, that still uses Unit_limit_groups heavily, i'll have to create the compatability manually.  But no big deal!  

I love the versatility of unit_type_tags, very cool!  

Great to hear R28 is coming soon!  I have another mini-project i'm working on for a couple bigger buttons (for pop up boxes and for scrolling through the city's 'what to build next' list) ----  for touch-screen or phone play of Civ 3.  ---- Now that SteamLink is out there and you can play Civ 3 on your phone!!!  

Should have it to you by this weekend, so maybe it can be part of R28 Preview 2.  AHAH!

PR #41, opened 2026-05-22
Allow configuration of unit view distance

### LINKED ISSUE #25: Configuration of terrain visibility
Hi, since fixing that bug I've opened up the repo again and started looking at things once more.
I'm considering how terrain visibility functions between different terrain types. I've not sighted anything regarding this already within C3X, outside of a couple mapped addresses used for current-visibility modifications.

So this issue functions both as feature request, and also as a request for information.

Are you aware of where this system lies? Earlier you mentioned sharing your ghidra repo - if developing this isn't immediately on the table for you but is still a reasonably achievable goal, is this option still on the table?

As for architecting an extension, I believe it can be boiled down to a couple of values per terrain type - notably a height, and then an adjustment that accounts for water granting additional visibility to adjacent tiles, and jungle preventing additional visibility to adjacent tiles. Finally, an adjustment from the unit can account for radar. Using larger numbers than the minimal and rounding, I think this can be extended for higher radius sightlines, though there should probably be some limit before the decision of which tile counts as intermediate makes everything appear wonky.

Cheers for your insights

Philiquaz:
Adds configuration of unit visibility ranges.
Adds configuration of terrain visibility behaviours.
Supports fortification bonuses, euclidean radii, the sea being flat, and all the bells and whistles.
The default configuration *should* match vanilla, but on a basis of caution everything is toggled under 1 flag. There is also a performance reason for this, as at a high visibility with hundreds of units around this can become a taxing system.
Resolves #25 

This was tested on the Steam version of the game.
Exe offsets should be correct for GoG as there is where most RE took place, but a test should be performed.

OFFSETS FOR PCG.DE ARE UNCHECKED, AS IS THE GAME.
UPDATE BEFORE MERGE.


Philiquaz:
Being clearer on my commenting, someone with the ready ability to check that game version would be helpful. The patches are all in locations calling can_see_tile and the update_visibility function

injected_code.c
```diff
@@ -1397,6 +1398,66 @@ parse_unit_type_limit (char ** p_cursor, struct error_line ** p_unrecognized_lin
...
+					rule.fortification_bonus_continent_lock = true;
+				} else
+					return RPR_PARSE_ERROR;
+			} else
+				rule.base_visibility = num;
+
+		} while (skip_punctuation (&cur, '+'));
+
+		int id;
+		if (find_unit_type_id_by_name (&name, 0, &id)) {
```
maxpetul: Multiple unit types may have the same name and in that case the rule should apply to all of them. That's not some rare edge case either, it's actually common for there to be duplicate types sharing the same name because of how the game handles types with different AI strategies. It splits them up into duplicate types so each type only has one strategy.

injected_code.c
```diff
@@ -1496,6 +1557,49 @@ read_recognizables (struct string_slice const * s,
...
+	int ival = 0;
+	if(!skip_white_space (&cur))
+		return 0;
+	for(int i=0; i<count; i++) {
+		if (parse_int (&cur, &ival) && (i == count-1 || (skip_punctuation (&cur, ',') && skip_white_space (&cur))))
+			array[i] = ival;
+		else
+			return 0;
+	}
+	struct string_slice sink;
```
maxpetul: This variable is not used. Same applies to read_fixed_bool_array.

default.c3x_config.ini
```diff
@@ -915,6 +915,42 @@ override_barbarian_activity_level_for_scenario_maps = none
...
+;                                [DESRT, PLAIN, GRSLD, TNDRA, FLDPL, HILLS, MNTNS, FORST, JUNGL, MARSH, VLCNO, COAST, SEA,   OCEAN]
+; The height a unit is considered to be standing at when on this type of tile.
+terrain_visibility_see_height =  [1,     1,     1,     1,     1,     1,     2,     1,     1,     1,     2,     1,     1,     1]
+; The hight a tile is considered to be seen from another tile. If this equals or exceed the see height, it will occlude tiles beyond.
+terrain_visibility_seen_height = [0,     0,     0,     0,     0,     1,     2,     1,     1,     0,     2,     0,     0,     0]
+; The bonus sight range granted by standing on or looking at a given tile
+terrain_visibility_bonus =       [0,     0,     0,     0,     0,     1,     1,     0,     0,     0,     1,     0,     0,     0]
+; Whether a unit on a tile with a bonus and seeing a tile with bonus is double rewarded. Otherwise, the large value is used. This may be necessary if you wish to use negative bonuses.
+terrain_visibility_bonus_can_stack = false
+; Whether tiles of the given type provide bonus view distance for seeing across them. This config only supports 0 and 1.
```
maxpetul: It's odd that the text says it "only supports 0 and 1" but the prefilled values are true/false.

default.c3x_config.ini
```diff
@@ -915,6 +915,42 @@ override_barbarian_activity_level_for_scenario_maps = none
...
+terrain_visibility_flat_bonus_limit = 1
+; Whether the flat bonus may stack with the regular, height-based bonus.
+terrain_visibility_flat_bonus_can_stack = false
+
+; A list of rules for individual unit types or unit classes (UTC_Land, UTC_Sea, UTC_Air)
+; A base visibility can be give, a multiplier to the height based bonus visibility can be give (times_bonus), and a single fortification bonus can be given.
+; There are two mutually-exclusive types of fortification bonus: when_fortified and when_fortified_same_continent. The latter only applies to tiles on the same continent.
+; The first rule matching a unit is used. Therefore, unit class rules should be at the end.
+; The following is an example: [Scout: 1 + 2 times_bonus, "Landmine": 0 + 0 times_bonus, "Army": 2 + 2 when_fortified]
+unit_visibility_rules = ["UTC_Sea": 2 when_fortified_same_continent]
```
maxpetul: I'd like you to change the format here a bit. First, `when_fortified_same_continent` should use hyphens instead like `when-fortified-same-continent`, same for the other keywords. Using hyphens instead of underscores here is arbitrary, but it's what I've been doing for all the other config keywords and I want to keep it consistent.

Second, I think it's a bit ugly to leak "UTC_" into an area users will see it. That prefix makes sense given that those values are part of an enum but it's strange without that context. I know the point of including the prefixes is so that the unit class names are less likely to conflict with actual type names, but I'd rather that be handled with quotation marks, as in e.g. `Sea` is a special word referring to naval units and `"Sea"` refers to a unit type named Sea. Come to think of it, `sea` should be accepted too as config keywords are generally not capitalized.

Edit: Also `unit_visibility_rules` should be `1 when_fortified_same_continent` by default to match the vanilla rules. This setting here annoyed me when testing because I was trying to set rules for UTC_Sea or Galley units in a scenario config and they weren't having any effect. I thought it was a bug at first before I remembered about this setting in the default config. It gets loaded first and overrides any settings for sea units in the scenario config. I think it's better to search for matching rules in reverse order so what's specified last takes precedence. That would be consistent with how other config options work.

C3X.h
```diff
@@ -483,6 +493,20 @@ struct c3x_config {
...
+	bool terrain_visibility_euclidean; //whether the upper bound should be determined by a euclidean metric than a chebyshev one.
+	int base_visibility_range; //should default to 1
+	int terrain_visibility_see_height[14]; //most tiles are 1, mountains+volcanoes are 2
+	int terrain_visibility_seen_height[14]; //most tiles are 0, forest+jungle+hills are 1, mountains+volcanoes are 2
+	int terrain_visibility_bonus[14]; //most tiles are 0, hills+mountains+volcanoes are 1
+	bool terrain_visibility_bonus_can_stack; //whether seeing hills and being on a hill provides double the bonus
+	bool terrain_visibility_flat_bonus[14]; //water tiles provide height bonus to tiles being seen across them [ie +2] / adjacent tiles always seen
+	int terrain_visibility_flat_bonus_limit; //maximum size of the flat bonus, in tiles
+	bool terrain_visibility_flat_bonus_can_stack; //whether flat bonus and regular bonus can both apply at once
+	struct unit_visibility_rule * unit_visibility_rule_list; //should default to naval unit type has +1 range in fortification with continent lock
```
maxpetul, line 592: This pointer should be freed in `reset_to_base_config`. Right now the memory is leaked.

maxpetul (changes requested):
Apologies for taking so long to have a look at this but I've just been busy with so many other things. It's good work. I was worried that this feature would be gnarly but this is a nice clean implementation. The changes I've requested are just small things. Also I've fixed the PCGames.de addresses and verified that it works on that EXE.

injected_code.c
Philiquaz: Fair enough, will see to it.
Does seem at a cursory glance some other features may need fixing?

default.c3x_config.ini
Philiquaz: Ah, I think I started off with 0/1 and then found the parser did support true/false but obviously forgot to change the doc

default.c3x_config.ini
Philiquaz: hyphens you are correct
I can handle mapping names to avoid UTC sure
reading in reverse order would make sense. it's a bit silly that these are grouped into the same list but it just ended up cleaner that way. Sea bonus itself is a bit odd because of the bonus distance over water tiles. Might need to re-check what matches vanilla but I recall the slightly unintuitive value because of this.

C3X.h
Philiquaz: good catch, will fix

Philiquaz:
No worries on the wait, there's a lot of stuff and I can fully understand taking time for a hobby.
Will get to the fixes soon(ish)
Thanks for patching up the pcgames.de addresses, it's not easy for me to work with that one :)

maxpetul:
Looks good except for two things. First, with the default rule `"Sea": 2 when-fortified-same-continent`, in my testing I see that giving fortified sea units four tiles of visibility over water, not three like under the standard rules. You said you tested it and settled on an unintuitive value, but can you test again? It makes sense that it would be four as: 1 base vis + 2 fortified same continent + 1 flat bonus over water.

Second, when I mentioned dropping "UTC_" from the Land/Sea/Air keywords in the config, I also had in mind that you would make it so that the keywords only applied if they were not in quotes. I.e., `Sea` in the config should refer to all sea units but `"Sea"` should refer to only the unit type(s) named `Sea`. So there should not be quotes in the default rule. It's probably unlikely that anyone will have unit types named Land, Sea, or Air, but we should leave open the possibility.

Philiquaz:
Okay, found the energy to actually look at the fortification code. Yes, it was unintentionally stacking with the flat bonus because the total sight distance was being recalculated, which I don't consider to be intentional. I've patched that and the config naming.

I would say the config of 2 is, although initially untuitive, correct. The mechanic works regardless of terrain or sightlines just like radar, so re-including terrain in its calculation is wrong. Relying on the flat bonus would incorrectly mean a tile at distance 3 obscured by land would not be seen when it should be.

Arguably if I spent more time I could configure fortification options to have more customizable stacking behaviour. Though then we're also looking at land height bonuses which is a realm the vanilla game doesn't touch.

maxpetul:
>The mechanic works regardless of terrain or sightlines just like radar, so re-including terrain in its calculation is wrong. Relying on the flat bonus would incorrectly mean a tile at distance 3 obscured by land would not be seen when it should be.

Alright, that makes sense.

I'd consider this PR finished and ready to merge at this point. There's just one very minor issue left which I was reminded of looking at your changes to the config: the grammar mistakes in "base visibility can be give... bonus visibility can be give", but I can fix that no problem. I'll merge this after PR #31 which should be very soon.

Philiquaz:
Bruh
Yes, intended to be "can be given" as in the 3rd clause.
Cheers for your work

PR #43, opened 2026-05-25
Feature/missile non lethality

drhaluu82-dev:
In Default (vanilla) Civ 3, if you set a unit as being a 'Cruise Missile', it will ALWAYS have lethal land/sea bombard, regardless of the Lethal Bombard Flags in the editor.  So it's impossible to make a Cruise missile non-lethal.  In the attached example, this Cruise Missile will always be land/sea Lethal, just based on its Cruise Missile flag, even though Lethal Land/Sea Bombardment isn't set.  
cruise1
cruise2

This feature (maybe even bug fix) makes the Lethal Land Bombardment and Lethal Sea Bombardment flags actually matter for Cruise Missiles (just like they matter for all other units).  

Config setting is "patch_cruise_missile_ignores_lethal_bombard_abilities" and I believe should be defaulted to FALSE, just in case a modder is unaware of the feature/bug fix.  They can turn it on manually by setting this config value to TRUE.  

**If testing, don't forget to set this config setting to True first!**

The code changes are highlighted in comments are "ToC-12" "ToC-13" or "ToC-13A".
ToC-12: Base code to ensure Lethal bombard settings in the editor matter for Cruise Missiles
ToC-13: AI adjustment to ensure AI does not waste missiles on 0 defense units that are out in the open (on a tile by themselves).  Seems this is Civ 3 vanilla hardcoded, that missile units cannot kill 0 defense units (ex. workers) that are out the open.  So this stops AI from attempting to fire missiles at such targets.  
ToC-13A: Adjustment to small bug found when creating ToC-12 that allowed first missile attack to skip a declaration of war.  this ensures that the warning pop up still occurs if attempting to bombard a previously allied civ with a missile, so that you must declare war first before engaging in bombard attack (Hidden Nationality ability still allows skipping of war declaration like in vanilla civ 3).  

**Missile Lethality functionality Tested and regression tested with vanilla civ 3 and the Tides of Crimson mod, works in game.**  

PR1

### DIFF
--- a/C3X.h
+++ b/C3X.h
@@ -407,6 +407,7 @@ struct c3x_config {
 	bool patch_division_by_zero_in_ai_alliance_eval;
 	bool patch_empty_army_movement;
 	bool patch_empty_army_combat_crash;
+	bool patch_cruise_missile_ignores_lethal_bombard_abilities;  // ToC-12 - Allow Missiles to obey normal lethal bombard flag rules
 	bool delete_off_map_ai_units;
 	bool fix_overlapping_specialist_yield_icons;
 	bool patch_premature_truncation_of_found_paths;
--- a/C3X.h
+++ b/C3X.h
@@ -78,6 +78,17 @@ struct unit_type_limit {
 	int cities_per;
 };
 
+/* ToC-26: Represents a named group of unit types that share a combined build limit.
+   Stored by pointer in unit_limit_groups (keyed by group name) and in unit_type_to_group
+   (keyed by unit_type_id).  The group struct is owned by unit_limit_groups; unit_type_to_group
+   holds non-owning pointers into it. */
+struct unit_limit_group {
+    int * unit_type_ids;        // Array of unit type IDs belonging to this group
+    int   count;                // Number of IDs in the array
+    struct unit_type_limit limit; // The shared limit value for this group
+    bool  has_limit;            // True once a limit has been assigned via unit_limits
+}; // END ToC-26
+
 struct work_area_improvement {
 	short improv_id;
 	int work_area_radius_limit;
@@ -320,6 +331,10 @@ struct c3x_config {
 	struct leader_era_alias_list * leader_era_alias_lists;
 	int count_leader_era_alias_lists;
 	struct table unit_limits; // Maps unit type names (strings) to pointers to limit objects (struct unit_type_limit *)
+	// ToC-26: Group-based unit limits. unit_limit_groups maps group name strings to struct unit_limit_group*.
+	// unit_type_to_group maps unit_type_id (int) to struct unit_limit_group* for O(1) runtime lookup.
+	struct table unit_limit_groups;
+	struct table unit_type_to_group; // END ToC-26
 	bool allow_upgrades_in_any_city;
 	bool do_not_generate_volcanos;
 	bool do_not_pollute_impassable_tiles;
@@ -1920,6 +1935,13 @@ struct injected_state {
 	} * penciled_in_upgrades;
 	int penciled_in_upgrade_count;
 	int penciled_in_upgrade_capacity;
+
+	// ToC-27: Set to true while patch_City_can_build_upgrade_type is running. When true,
+	// patch_City_can_build_unit skips its unit-type limit check so that upgrades to group-limited
+	// types are not blocked at the City_can_build level. patch_Unit_can_upgrade re-applies the
+	// limit with full source-type context, allowing same-group upgrades (net-zero count change)
+	// while still blocking over-limit production and cross-group upgrade attempts.
+	bool checking_upgrade_type_eligibility; // END ToC-27
 
 	// While in Leader::do_capture_city, the city in question is stored in this var. Otherwise it's NULL.
 	City * currently_capturing_city;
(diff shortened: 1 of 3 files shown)


maxpetul:
This doesn't quite work. Your edit to `patch_Fighter_find_actual_bombard_defender` stops non-lethal cruise missiles from killing 1-HP units by preventing those units from being selected as targets for damage. However that only fully prevents lethal damage when cruise missiles have a rate-of-fire of 1. Otherwise, a unit with 2 or more HP could be selected as a target then killed by multiple hits.

Another issue is that much of the code you've inserted won't run because it's part of the regular bombard logic which is not used for cruise missiles. Cruise missiles have their own mostly separate code path. When `Unit::bombard_tile` is called for a cruise missile, it calls `Unit::begin_bombarding_tile` like always, then that method checks whether the unit is a cruise missile and, if so, calls `Unit::strike_tile_as_cruise_missile` and after returns false so the rest of `Unit::bombard_tile` is skipped. That means `Fighter::do_bombard_tile` is never called for cruise missiles. Your original approach of clearing the cruise missile flag in `begin_bombarding_tile` would have avoided this issue, but I wouldn't have liked that approach either as it would cut out the entire cruise missile logic and have made missiles use the standard bombard logic instead. That's a large unnecessary change in my opinion.

To make cruise missiles respect lethal bombard abilities, this should be all you need:
- Replace the hardcoded true/true for the `land_lethal`/`sea_lethal` params passed to `find_defender_against_bombardment`. That function is called from four places. The two places where the game calls it to get the defender before applying damage are already redirected to `patch_Fighter_find_actual_bombard_defender`, you've handled those, that's fine. The other two calls are from `Unit::ai_eval_bombard_target`, which is part of the regular bombard logic and not used for cruise missiles, and from `Unit::ai_eval_cruise_missile_target`, which is of course the one you want to modify. Modifying that call should work to make the AI aware of this feature instead of your edit to `patch_Unit_ai_eval_bombard_target` (which won't run anyway because that's only for regular bombardment not cruise missiles). By the way, the way that the params for the lethal flags are hardcoded to true here makes me think it's not a bug that cruise missile strikes are always lethal. That looks like a deliberate choice.

- Modify `can_damage_bombarding`, although that function already respects the lethal land & sea flags for cruise missiles since I didn't know they were ignored when I wrote it. Still, it should be modified to be consistent with everything else. That ensures that stack bombard and stealth attack bombard work properly.

- Somehow modify the loop that applies damage after a cruise missile strike to stop before the target is killed in case the strike is non-lethal. This is the difficult part. The relevant method here is `Fighter::do_cruise_missile_strike`. I think the easiest approach would be to modify the call to `rand_int` that rolls to do damage on each round so that damage never happens when the target has 1 HP remaining and the cruise missile strike is not lethal. You'd have to drag that information in from elsewhere, probably easiest to intercept the earlier call to `check_combat_anim_visibility` since it's passed the cruise missile unit and the defender as parameters.

Do you have the Ghidra project? I forget who I've sent it to. If not, you won't be able to bring up `do_cruise_missile_strike` or find that other call to `find_defender_against_bombardment`. I'll help you out with that if you need it, it wouldn't take much time.

drhaluu82-dev:
@maxpetul thanks so much for the detailed explanations, Max!  Very helpful & you are absolutely correct.  Your explanation also helped me realize that this feature might not be used by many people at all.  It was needed in my Tides of Crimson mod (in development) to ensure certain one-time use "Items" that were lower-tier were differentiated from higher tier "Items".

Let me give it a good ol' college try again when i get that chance, and see if I can fix this using your guidance.  

I don't think i have the Ghidra project, would love a little bit of help if you don't mind.  I'm still learning so would love to get more knowledge.  

Nothing urgent of course, please take your time.  I'm sure this small feature is not really being asked for and is not needed in R28,  I initially thought this feature would be more useful than it actually is (assuming I can get it working).   =D 

Thanks again for all your help!

maxpetul:
Sure thing, I'm glad to help. If you want the Ghidra project, PM me on CivFanatics. Ghidra itself is somewhat complicated and takes time to learn to use. It may or may not be worth the trouble for you, it depends on how much you're planning to contribute to C3X in the future. I'll gladly send you the project if you want it, though.

drhaluu82-dev:
@maxpetul sorry for the delay in response!  I've been studying for a certification exam I have to take in about a month!  

If it doesn't hurt then yes please send, I'm always willing to learn.  C in general is brand new to me.  Would love to see what it's all about!  

I think I might have found a way to get a non-lethal missile working, without using Ghidra actually, will need to play around a little bit more.  
Will try to send something this weekend.  

And yes, my goal is to consistently contribute a new feature to C3X every 2-4 weeks or so.  I might be out of pocket for a while due to studying for this exam, but once it's over I do want to get back in the horse!  I have a Buildings limit feature and an Anti-road spaghetti feature I want to add as well eventually, among others!  

I've been working, when I get a chance, on the Tides of crimson mod, and a lot of code is very niche to the mod or just ugly code that I'm embarrassed to send over to you, but some things that might be useful that I can find a way to clean up I'll try to send!  

I love C3X and all the great work you guys have done in this so far!!!

PR #44, opened 2026-05-25
Bug fix: Keep generated resource tile state valid across recompute redraws

instafluff0:
This is a fix for som_rybov's bug report, [here](https://forums.civfanatics.com/threads/c3x-exe-mod-including-bug-fixes-stack-bombard-and-much-more.666881/post-16953043), save attached.

In a nutshell, I traced the crash back to `Tile_has_colony` getting a NULL tile. However patching that (bailing if NULL) fixed the symptom but not the underlying problem. 

As best I can tell, the crash is from `Trade_Net::recompute_resources` receiving a bad tile pointer while we have the temporarily expanded Map.TileCount to include generated resource tiles.

1. `patch_Trade_Net_recompute_resources builds` is->resource_tiles.
2. It saves the real tile count:
```c
is->saved_tile_count = p_bic_data->Map.TileCount;
p_bic_data->Map.TileCount += is->count_resource_tiles;
```
3. `Trade_Net::recompute_resources` loops over 0 .. Map.TileCount - 1.
4. The patched `Map_get_tile_when_recomputing_resources_*` functions are supposed to return:
  - real map tile if index < saved_tile_count
  - generated resource tile otherwise
5. But `Tile::has_colony(NULL)` means (I think) the first tile lookup returned NULL.

So if a resource recompute happens while saved_tile_count is already active, then saved_tile_count can be overwritten with an already-inflated tile count. Then an artificial tile index is misclassified as a real map index, so the wrapper calls real Map_get_tile with an out-of-range index. That returns NULL.

The saved game moves to the next turn after these changes, and calculated resources appear correct afterward.

Also: I noticed in the course of debugging this we had duplicate `define Tile_has_colony` entries in civ_prog_objects.csv and removed the 2nd one.

PS - holy moly, the repo is really heating up with PRs. I hope all is well and this isn't all too much too quickly.

[bug.zip](https://github.com/user-attachments/files/28231512/bug.zip)

### DIFF
--- a/C3X.h
+++ b/C3X.h
@@ -934,7 +934,7 @@ const struct district_config special_district_defaults[USED_SPECIAL_DISTRICT_TYP
 		.buildable_square_types_mask = DEFAULT_DISTRICT_BUILDABLE_MASK,
 		.img_path_count = 5, .img_column_count = 4, .btn_tile_sheet_column = 0, .btn_tile_sheet_row = 0,
 		.culture_bonus = 1, .science_bonus = 1, .food_bonus = 0, .gold_bonus = 1, .shield_bonus = 0, .happiness_bonus = 0, .defense_bonus_percent = 25,
-		.generated_resource = NULL, .generated_resource_id = -1, .generated_resource_flags = 0
+		.generated_resource = NULL, .generated_resource_id = -1, .generated_resource_flags = 0, .auto_add_road = true
 
 	},
 	{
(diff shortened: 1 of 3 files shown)


maxpetul:
Thanks for the fix. This all looks fine except for the deletion of line 24930, why'd you do that? Also seeing this reminds me I wanted to combine all those `patch_Map_get_tile...` functions. There's no reason to have separate patch functions if they're all the same.

instafluff0:
> This all looks fine except for the deletion of line 24930, why'd you do that?

I just re-added that and indeed som_rybov's crash doesn't happen, so I guess in this case (or all? I added it defensively) it is not necessary. Honestly I just re-reviewed and am not sure why I removed that line. I recall reasoning that if `saved_tile_count` was -1, the `Map_get_tile` functions would no longer be able to distinguish a real tile index from resource-tile ones, which I thought may be leading to the null tile pointer.

That was as I was trying to work through the crash call stack and think through what the issue could be, and talking it through now I agree it doesn't seem necessary. And of course the save works even with that line. Let me re-add it.

maxpetul:
Alright, that makes sense. The program would work whether that line is there or not. I like having it there because it clears the saved tile count variable after the count has been restored, so the true tile count is only ever in one place and you can know where it is by checking `is->saved_tile_count >= 0`. Merging now...

PR #46, opened 2026-06-19
R28 seasonal cycle no cache

instafluff0:
As discussed, here is the no-cache version of seasonal + day-night cycle, mostly Codex on this one.

I've tested with all varieties of settings, including going to main menu and starting new games or loading saves, etc.

One other thing to mention: after the initial pass, I observed mem usage going up ~20MB per turn when `every-turn` was true for either cycle. After poking around Ghidra, it turns out that adding the line:

```c
if (img->JGL.Image != NULL)
    img->vtable->clear_JGL (img);
```

in `read_in_dir()` is critical, as otherwise we would have had a major memory leak. Looking at this more deeply, I *think* we actually had this before as well, it just wasn't apparent because (1) loading of images was generally a 1-time thing and (2) we knew those images took a lot of memory, so this additional unnecessary usage was not apparent. It became obvious though when we stopped caching.

Anecdotally, that background processing I mentioned after closing the game? Seems to be gone, or at least ends very quickly, after adding this.

Edit: I also happened to find that we were hard-coding the resource counts for cycle images, so updated that code to make that dynamic as well. Looks good in testing.


instafluff0:
As an aside, I have a branch ([instafluff-current](https://github.com/instafluff0/C3X_Districts/tree/instafluff-current)) which includes my PRs with the main repo merged, for testing and fun. I also had time to finally put my [long-intended scenario](https://github.com/instafluff0/instafluff_scenario) together, which is basically the standard game, but with districts, tile animations, seasons, some additions to the modern tech tree & units (adds silicon wafers & other resources, AGI as a tech & Wonder, autonomous drones, space elevator etc.) and some extra wonders added in for flavor:

image

image

test2

Anyway I'm somewhat amazed that it all works and came together - and so much of this is thanks to you. I've got to say it feels like a whole different game. I know I've said it before and we still have a couple PRs to get through, but seriously, thanks.

maxpetul:
You're very welcome. It's incredible what you've been able to do and doubly so how quickly you've managed to do it. I'll try to review your last few PRs sooner rather than later. I'm aware that I'm a bit of a ball and chain on your very rapid progress. It's just every time I sit down to look into those PRs, something else comes up, sometimes related to the mod and sometimes real life stuff.

instafluff0:
Regarding the PRs, please take your time! Selfishly, given that my dream-features are all working in the branch I mentioned, it's not at all a big deal to me if merging them takes longer. Even later in the summer/fall would be fine. I think it's more important that you get to enjoy it and work on fun features, rather than reviewing PRs all day.

maxpetul:
By the way, there's a bug in your patch to `Leader::get_attitude_toward`. The relations penalty for extraterritorial colonies should be added to, not subtracted from, the attitude because higher values mean worse relations. I discovered this (speaking of distractions) when I was investigating how the AI decides when to declare war and who against. I was sent down that rabbit hole a few days ago after someone emailed me asking how the AI aggression setting works and whether the AI unfairly targets the human player. That brought me to the function that converts the AI attitude in points into diplomatic mood. It's at 0x440AD0, Leader vtable method m34, I named it `get_diplomatic_mood_toward`. I checked in game that it returns 0 for gracious, 1 for polite, and so forth, and given that it's clear that higher values from `get_attitude_toward` correspond to more hostility. Also if you read `get_attitude_toward` itself, you'll see that for example being in a shunned government adds 5 to the result whereas favored govt subtracts 5. So that's what I've been doing recently instead of reviewing your PRs, sorry. Once I've worked out all the rules I'll post a thread about this to CivFanatics since I expect many in the community will be interested beyond just the guy who emailed me.

instafluff0:
Huh, wow. Sorry for not catching that and getting it backwards.

Just a one-line fix, right? Want a separate PR for that, or should I add in here? Or are you already taking care of it?

maxpetul:
Yeah, it's a one line fix. I took care of it here: https://github.com/maxpetul/C3X/commit/dc26d764baf608d8edd687a963da70c2e122b202

maxpetul:
Alright, finally had a look at this. It looks good to me except I just wish you'd gone further and completely flattened the things that used to be multiplied by hour and season. In particular:
- There is no need for `insert_spritelist_proxies`, `insert_sprite_proxies`, and `insert_sprite_proxy` to take `season` and `hour` params anymore.
- There is no reason to allocate `day_night_sprite_proxy_by_season_and_hour` dynamically any more, and by making it static you could delete a bunch of NULL checks.
- `day_night_sprite_proxy_by_season_and_hour` shouldn't be called ...by_season_and_hour any more since entries are no longer looked up by season and hour. It's now just a table that maps sprite pointers from the Map Renderer to the corresponding sprite ptrs in `cycle_imgs`. Not sure what a better name would be.

Another small thing that bothers me is the comment "The day/night cycle sprite proxies will have been cleared..." in `patch_move_game_data`. It's now just floating there not attached to any code. I'm not sure it's needed at all, but if you want to keep it, it should be placed above the if statement calling `reload_current_day_night_and_seasonal_images` at the end of that method, noting that the images must be loaded there because `Map_Renderer::load_images` isn't necessarily called when a game is loaded.

instafluff0:
Thanks for reviewing and you're right about the cleanup stuff. I believe I've addressed all 4 of these now in the last commit, and also played through a few cycle changes and confirmed everything continues to work as expected.

maxpetul:
Great, thanks.

PR #49, opened 2026-06-27
Distro hub city select

instafluff0:
Another PR, sorry. There is *zero* rush on this at all - as far as I know only one person has ever asked for it.

This allows the user to set which cities receive distro hub benefits.

Note this adds an inlead for `0x61B320`, which I tracked down to being a ContextMenu call, which I named `Context_Menu_draw_item`.


instafluff0:
Closing in favor of https://github.com/maxpetul/C3X/pull/50, which doesn't have the baggage of all these other features.

PR #51, opened 2026-07-07
Allow pollution spawns to alternatively reduce population instead of …

instafluff0:
A minor feature: I find pollution useful in the sense that it somewhat counterbalances breakaway industrialization, but like others find cleaning it up with workers very tedious. This alternatively allows it to simply reduce population by 1, instead of polluting a tile. It leverages the same code we use elsewhere for calculating pollution with a larger work area.


Philiquaz:
Perhaps a 2nd flag so that _both_ is an option?

instafluff0:
Sure, though I made it an enum rather than 2nd flag entirely.

maxpetul:
Looks fine to me. Just one tiny thing: the option to leave the game rules unchanged should have been called "standard" instead of "default" for consistency with a few other config options.

PR #52, opened 2026-07-15
Allow exemption of losing certain units contributing to war weariness

instafluff0:
Another minor feature, and zero rush on this: allow users to prevent the loss of certain units from contributing to war weariness. In my personal scenario, this is really about drones, robots and so on, and adds a touch of modernity & realism. It also interestingly makes late-game long wars with Democracies, etc. viable to some extent.

Note that the separate `repl call`s and the need for `is->war_weariness_fight_context` in `Fighter_fight` are necessary as that is the only function I found where both the attack and defender are assessed for addition of war weariness (if HP drops to 1).


instafluff0:
~~Possibly found a bug, closing till I sort that out.~~

Fixed, tested and working as expected.

instafluff0:
I recalled that surviving units with 1 remaining HP also cause war weariness, will patch those cases as well and reopen...

instafluff0:
Okay, additional HP of 1-related war weariness cases patched, updated, and tested. Reopening...

PR #54, opened 2026-07-18
Ensure tradenet recalcs due to auto-adding roads/rails always see true tile count

instafluff0:
Another bug fix. I thought I had slain the monster caused by auto-adding roads when the tile count was artificially inflated, but it rose again. I have a deterministic save that crashes in the interturn each time that I'm happy to provide to test with, if you'd like (with [my scenario](https://forums.civfanatics.com/threads/instafluffs-scenario.703788/) though, so you'd need that too).

The solution I settled on is just to replace the direct `tile->vtable->m56_Set_Tile_Flags` call with a small wrapper function, `district_set_tile_flags_safely`, that ensures that when Trade Net side effects are run due to road/rail additions, the true tile count is always the one present.

If you want me to rename the function just let me know, or feel free to.

### DIFF
--- a/injected_code.c
+++ b/injected_code.c
@@ -6786,6 +6786,22 @@ pick_missing_district_for_improvement (City * city, struct district_building_pre
 	return fallback;
 }
 
+void
+district_set_tile_flags_safely (Tile * tile, int layer, int flags, int tile_x, int tile_y)
+{
+	int restore_tile_count = -1;
+
+	if ((is->saved_tile_count >= 0) &&
+	    (p_bic_data->Map.TileCount != is->saved_tile_count)) {
+		restore_tile_count = p_bic_data->Map.TileCount;
+		p_bic_data->Map.TileCount = is->saved_tile_count;
+	}
+
+	tile->vtable->m56_Set_Tile_Flags (tile, __, layer, flags, tile_x, tile_y);
+
+	if ((restore_tile_count >= 0) && (is->saved_tile_count >= 0))
+		p_bic_data->Map.TileCount = restore_tile_count;
+}
 
 bool
 district_is_complete(Tile * tile, int district_id)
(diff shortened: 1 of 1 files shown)


maxpetul:
Feels to me like the monster may still not be fully vanquished, but maybe you understand this bug better than me. Here's my understanding and correct me if I'm wrong:
1. `Trade_Net::recompute_resources` gets called and replaces the tile count to enable resource generation.
2. Somewhere inside `recompute_resources`, `district_is_complete` gets called.
3. If the district was just completed and auto-adds a road, `district_is_complete` calls `m56_Set_Tile_Flags` to add a road.
4. `m56_Set_Tile_Flags` crashes if it's called to add a road while the tile count has been manipulated. Likely because it calls `Trade_Net::recompute_network_around`, I can see that method uses the tile count.

I'd like to know how `district_is_complete` gets called from `recompute_resources`, is that something you've figured out? I'm wary about restoring the tile count early unless we're sure that the loop over tiles is finished. A simpler solution than what you've done, and preferable in my opinion, would be to make it so that districts don't become complete while the tile count has been modified for recomputing resources. I.e., place `if (is->saved_tile_count >= 0) return false;` in `district_is_complete` before `// Mark as completed and run one-time side effects`. That ought to work, right? Do you know if that would cause any problems? Do districts need to be completed at that point?

The proper solution here would be to obviate the tile count replacement by patching the loop end condition for the loop over tiles in `recompute_resources`. The only reason I didn't do that in the first place was that it was easier to modify the tile count instead to get that loop to continue for longer. But if it's causing a bunch of other problems, maybe it's not actually easier after all.

instafluff0:
> I'd like to know how district_is_complete gets called from recompute_resources, is that something you've figured out?

Not completely. I've tried setting `p_OutputDebugStringA` at each function I expected to get called and so on, but haven't quite cracked exactly why this happens as I really only seem to be seeing `district_is_complete` over and over in the output. 

- From the debugger, I can see the actual line that errors out is indeed in `Trade_Net::recompute_network_around`, specifically the `if` call on the line after `Map::get_tile(&bic_data.Map,iVar7);` toward the end. Presumably the tile is null.
- The function call preceding that in the stack trace is `m56_Set_Tile_Flags`.
- Before that: `district_is_complete`
- Before that: I mostly can't tell the next few based on the recompiled code, though the 11th call down is definitely `Trade_Net::recompute_network_around` again, and a number of functions look recursive and are called multiple times in a row.

Also, following your suggestion I tried alternatively using 

```c
...
if (is->saved_tile_count >= 0) 
	return false;

// Mark as completed and run one-time side effects
inst->state = DS_COMPLETED;
inst->completed_turn = *p_current_turn_no;
...
```

in `district_is_complete`, ~~but the result appears to be an infinite loop, with `district_is_complete` getting called over and over again. I'm with you in thinking that's a cleaner solution, but for reasons that are unclear to me deferring the district completion side effects don't seem to solve it.~~

~~As to whether using a different strategy than the tile count replacement is the way to go, I'm not sure. Clearly it all worked fine until I added in this silly feature for districts adding roads/rails, so it seems a waste to have to revamp it just for that.~~

---

**Edit**: Nevermind, I spoke too soon! This was dumb - I'm on a huge map with many districts, and with the debugger doing so much the turn just took a long time and there were many `district_is_complete` calls, not an infinite loop.  

`if (is->saved_tile_count >= 0) return false;` in `district_is_complete` appears to actually work just fine.

That's a relief. I'll update the PR to just do that.

maxpetul:
>Not completely. I've tried setting `p_OutputDebugStringA` at each function I expected to get called and so on, but haven't quite cracked exactly why this happens as I really only seem to be seeing `district_is_complete` over and over in the output.

That's too bad since it means we can't confidently say that there are no other possible crashes caused by this same issue. This is already the second one. The first happened when `recompute_resources` triggered the first time resource connection popup while the tile count was manipulated. That caused the map to be redrawn which crashed with an invalid tile count. I fixed that by restoring the tile count in `patch_Map_Renderer_m71_Draw_Tiles` when necessary. Anyway, this is fine for now.

PR #55, opened 2026-07-22
Ensure if governor is on, worked tiles & specialties are optimized after district yields are added in

instafluff0:
In cases where city governors are managing tiles/specialists, district & distro hub benefits can be recalculated/redistributed in the interturn (for example, reducing food) and the governor management function isn't automatically run afterward. The result is that often at the start of turns, a city with a governor active can appear to be starving, etc., and the user needs to manually open the city view, open the governor window then close it, in order for the governor management process to be run with district/distro hub yields factored in.

The changes here just ensure that whenever district/distro hubs yields are recalculated for a city, we run the governor process (if activated; I'm lazy so always leave gov. on) afterward to avoid needing the user to manually run this for each city.


maxpetul:
Alright, looks good.

PR #59, opened 2026-08-16
Display combat win rate

bingyu-893:
### Overview

This PR implements a real-time combat win rate display (inspired by Civilization IV) that dynamically computes and displays combat win probabilities directly in the UI.

---

### Key Features & Implementation

* **Combat Display Modes:**
* **`Compact` (Default):** Displays only the final calculated win rate for a clean, minimal UI.
* **`Detailed`:** Shows the win rate alongside the attacker and defender combat modifiers. To prevent screen clutter, it only displays the final effective modifier and whether defensive bonuses are ignored, rather than listing every individual modifier.

* **Bombardment Hit Chance:**
* Displays the probability of dealing **at least 1 HP of damage**.
* *Design note:* Rate-of-fire distributions were intentionally simplified to keep the tooltip readable and avoid cluttering the interface.

---

### UI Behavior 

Currently, the odds display box remains **persistent on the screen** while targeting/active. I initially tested having the overlay appear strictly when hovering over an enemy unit. However, during actual playtesting, the constant pop-in and UI flickering felt visually disruptive.
* **Feedback needed:**
Do you prefer the persistent box, or would you rather stick to hover-only? Or alternatively, should I expose this as a user-configurable toggle in the settings/config?


maxpetul:
I'd prefer that the box only appears when hovering over a target. Hopefully the UI flickering can be fixed somehow. In the mean time, you might as well add a toggle for persistence so users can switch that to deal with the flickering in case it's too annoying and too hard for us to solve.

Overall, this PR looks good. There are just a few issues I found:

- When there are 20 or fewer turns left in a game, the game shows a little box in the lower left of the screen above the minimap reading "Turns Left: XX". Right now, this overlaps with the win rate display, which is drawn in the same area:
turns_left_box_obscured<br>Easiest solution is probably to move the win rate box upward when there are <= 20 turns remaining.
- When the double sized minimap is active, the win rate box is drawn above it aligned to its right edge:
box_right_aligned_above_minimap<br>I think it would look better if it were aligned to the left edge of the screen.
- The background images must be deinitialized whenever a new scenario is loaded because they might be replaced with custom versions for a particular scenario. This is done for all art files that can be customized, see the comment "// This scenario might use different mod art assets than the old one".

maxpetul:
Perfect. Thanks for the contribution.

PR #60, opened 2026-08-17
Add Municipal District

instafluff0:
This is a district that ZergMazter has been coordinating with me and working on the art for. This adds the district art, button art, and config entry for it. This is also the first district to use the `render_strategy = by-building` flag, so the buildings themselves don't need to be dependent on one another.

image

Screenshot 2026-08-17 at 3 04 12 PM

**Edit**: I realized I still need to do day/night annotations for this. Please hold off on merging until I follow up.


instafluff0:
Update: please feel free to merge as-is if you're ok with it. I'll get to the day-night work with it when I can.

PR #61, opened 2026-08-24
Fix custom tile improvements used by AI & include bonus yields

instafluff0:
A few users pointed out the districts of `type = tile-improvement` weren't being automatically worked after being built, and also that their yields didn't include that of bonus resources (or extra shield grassland tiles, for that matter). i.e., they weren't really behaving exactly like mines or irrigation would.

This fixes both of those cases and looks good in my testing. 

Also, please note that I took a close look at usage in Ghidra and am pretty sure `Map_calc_shield_yield_at` should have exactly the same arguments/signature as  `Map_calc_food_yield_at`. I found that as plugging in the previous `City * city` argument (the 4th one, I think) crashed the game, while using the final argument (`int param_6`) seemed to work correctly. Looking closely at usage of the function in Ghidra seemed to confirm this.


maxpetul:
This looks good to me too. You're right about `Map::calc_shield_yield_at`. I don't know why I had its signature set incorrectly, I think I just never paid any attention to it.

Can you elaborate about the fix for the fact that improvement-type districts "weren't being automatically worked after being built"? I'm wondering what to write in the changelog. If I write that it was fixed that they were not automatically being worked, that implies that now they are automatically worked, but that's not necessarily the case, right? Is it that the governor now may set a citizen to work them?

instafluff0:
> Is it that the governor now may set a citizen to work them?

Thanks! Yes, that's exactly right. The governor would never set a city to actually work the tile after the improvement was built.

PR #62, opened 2026-08-24
Consolidate district port/aerodrome unit checks and add additional resource & obsolence logic

instafluff0:
A user pointed out that ports were enabling obsolete naval units to be built, which I realized also likely was an issue with air units. After reviewing the code, I noticed we also weren't checking for required resources in the given city.

I consolidated the logic a bit and added the resource and upgrade checks, including the special logic for units triggering golden ages from vanilla. Last, I noticed the code that read as:

```c
// Superficially allow the AI to choose the unit for scoring and production.
// If a disallowed air/naval unit is chosen in ai_choose_production, we'll swap it out for a feasible fallback later
// after prioritizing the aerodrome/port to be built
if (! is_human && (
    (type->Unit_Class == UTC_Air || city_can_build_district (this, AERODROME_DISTRICT_ID)) ||
    (type->Unit_Class == UTC_Sea || city_can_build_district (this, PORT_DISTRICT_ID)))
)
```

... should read as `(type->Unit_Class == UTC_Air && city_can_build_district (this, AERODROME_DISTRICT_ID))`, not `||`.

---

**Edit**: it's been pointed out to me that naval and air units can't upgrade currently with ports and aerodromes and the `*not_cities` flags on. I'll work on that next, please hold off on merging for now.

### DIFF
--- a/injected_code.c
+++ b/injected_code.c
@@ -25116,11 +25116,6 @@ patch_City_can_build_unit (City * this, int edx, int unit_type_id, bool exclude_
 		bool is_human = (*p_human_player_bits & (1 << this->Body.CivID)) != 0;
 		UnitType * type = &p_bic_data->UnitTypes[unit_type_id];
 
-		// Bail if tech reqs are not met
-		int prereq_id = type->AdvReq;
-		if (prereq_id >= 0 && ! Leader_has_tech (&leaders[this->Body.CivID], __, prereq_id))
-			return false;
-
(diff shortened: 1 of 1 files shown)


maxpetul:
How's this coming along? I'd like to put out the final R28 within a week or two including fixes for all known bugs, so I'd like to merge this PR before then if possible.

instafluff0:
Sorry about the hold up. I have not had time to work on this lately. If you're ok with this being a known bug I'd say release without it. I'm not sure if I'll have free time to address it in the short-term.

maxpetul:
Alright, no problem.

